PR #34, opened 2014-08-18
Added more descriptive text to the Slack config page.

jonathancadepowers:
I added text to the interface making it clear that you must enter # before the channel name. If you fail to do so, you'll get a 500. This came up via a recent support case.


brynary:
This is trivial, so merging it.

PR #43, opened 2015-01-13
Added more descriptive wording for Slack channel config.

jonathancadepowers:
Users sometimes fail to add a `#` character before the channel's name, which breaks the configuration. I added wording to make this clear.

### DIFF
--- a/lib/cc/services/slack.rb
+++ b/lib/cc/services/slack.rb
@@ -9,7 +9,7 @@ class CC::Service::Slack < CC::Service
       description: "The Slack webhook URL you would like message posted to"
 
     attribute :channel, String,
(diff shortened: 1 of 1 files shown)


JacobNinja:
This still doesn't solve the problem of the user not providing `#` as a predicate to the channel name.

I think we should detect the presence of `#` and add it if not present. I'm happy to walk through and discuss the process of writing this code with you.

noahd1:
Agree with @JacobNinja and think someone should write the code instead of changing the wording. After that, the wording should specify that it's optional (otherwise people will wonder)

brynary:
This seems to have gone stale. Closing.

@jonathancadepowers feel free to re-surface as desired.

PR #44, opened 2015-01-19
Fix type inference for Password type

sikachu:
The original type inference[1] uses `equal?` which returns `true` when any subclass, including `String`, is passed in. This fixes an issue when a `String` type is mistakenly marked as `Password` type when passed in.

[1]: http://git.io/SXJanA

### DIFF
--- a/lib/axiom/types/password.rb
+++ b/lib/axiom/types/password.rb
@@ -1,1 +1,7 @@
-class Axiom::Types::Password < Axiom::Types::String; end
+class Axiom::Types::Password < Axiom::Types::String
+  def self.infer(object)
+    if object == Axiom::Types::Password
+      self
+    end
+  end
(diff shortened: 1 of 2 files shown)


brynary:
Can we get a test for this?

sikachu:
@brynary We sure can! Let me port this [vague test case](https://github.com/codeclimate/codeclimate/blob/842240499d376d8bd5c4619337ef3147700ef348/spec/models/service_spec.rb#L83-L88) from codeclimate/codeclimate to here.

sikachu:
@brynary I've updated the PR with test. Let me know what you think.

sikachu:
As discussed in Slack, we'll roll out this change on codeclimate/codeclimate first, see if nothing breaks, then merge it.

PR #46, opened 2015-01-19
Cleanup gem specifications

sikachu:
* Remove `bin` from executables, as our gem actually doesn't have any executable. Those stuff in `bin` are binstubs. That confused RubyGems and conflicted executable in user's installation.
* Relax Bundler dependency as current version of Bundler is 1.7.x.


calavera:
:+1:

PR #47, opened 2015-01-19
Update Travis configuration and add test-unit gem

sikachu:
* Add Ruby 2.0 and 2.2 to build matrix.
* Use Travis' container infrastructure.
* Add `test-unit` gem because Ruby 2.2 requires it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,12 @@ language: ruby
 cache: bundler
 rvm:
   - 1.9.3
-  - 2.1.2
+  - 2.0
+  - 2.1
+  - 2.2
+sudo: false
 env:
   global:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,9 +2,12 @@ language: ruby
 cache: bundler
 rvm:
   - 1.9.3
```
jeffrafter: Are we ready to remove 1.9.3?

sikachu: It doesn't hurt to leave it in _for now_, as EOL is Feb 27. We can remove it after that.

sikachu: Also, codeclimate/codeclimate runs on 1.9.3. We might want to keep it around until we're on 2.x bandwagon.

PR #48, opened 2015-02-03
Use more accurate testing of OAuth Token for PR Service

noahd1:
Github returns a `422` if you post to the statuses API with a valid token but invalid SHA.

Github returns a `401` if the API token is invalid.

/cc @brynary


noahd1:
Closing in favor of #49 which removes bundler-dependency related commit.

PR #50, opened 2015-02-04
Check config of GH PR service and test appropriately for status/comment ability

noahd1:
Based on the configuration of the GH PR service, the "test" will check the permissions of the token to comment or post a status.

I attempted to use the same "no-op" API call style check for comments, but it didn't work in the same way: the API returns a 404 in both the case that the token has insufficient scope as well as the case the the case that the PR number doesn't exist (PR #0 is what I attempted to use, other PRs may not exist yet).

The alternative solution -- checking the headers for x-oauth-scopes -- is pretty good. However, I'm also not certain that it's a definitive solution: I believe there's at least one scenario -- a "machine user" with read-only repo access -- where you can have the proper header (with "repo" scope) but still not be able to post a comment. 

/cc @brynary

### DIFF
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -43,6 +43,8 @@ class CC::Service::GitHubPullRequests < CC::Service
   # additional information (github-slug, PR number, etc) we can't test much
   # else.
   def receive_test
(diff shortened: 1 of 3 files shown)


calavera:
It looks good. :+1:

PR #52, opened 2015-02-12
Services return the HTTP call's status

jyurek:
In order to properly record the data for ServiceEvents in codeclimate
proper, we need to have the calls to the external services return
something useful. This is the first step toward that goal.

### DIFF
--- a/lib/cc/service/http.rb
+++ b/lib/cc/service/http.rb
@@ -15,11 +15,11 @@ module CC::Service::HTTP
     end
   end
 
-  def get(url = nil, body = nil, headers = nil, &block)
+  def service_get(url = nil, body = nil, headers = nil, &block)
     raw_get(url, body, headers, &block)
   end
 
-  def post(url, body = nil, headers = nil, &block)
+  def service_post(url, body = nil, headers = nil, &block)
     block ||= lambda{|*args| Hash.new }
     response = raw_post(url, body, headers)
     {
--- a/lib/cc/service/http.rb
+++ b/lib/cc/service/http.rb
@@ -15,7 +15,23 @@ module CC::Service::HTTP
     end
   end
 
-  def http_get(url = nil, params = nil, headers = nil)
+  def get(url = nil, body = nil, headers = nil, &block)
+    raw_get(url, body, headers, &block)
+  end
+
+  def post(url, body = nil, headers = nil, &block)
+    block ||= lambda{|*args| Hash.new }
+    response = raw_post(url, body, headers)
+    {
+      ok: response.success?,
+      params: body.as_json,
+      endpoint_url: url,
+      status: response.status,
+      message: "Success"
+    }.merge(block.call(response))
+  end
+
+  def raw_get(url = nil, params = nil, headers = nil)
     http.get do |req|
       req.url(url)                if url
       req.params.update(params)   if params
@@ -24,7 +40,7 @@ module CC::Service::HTTP
     end
   end
 
-  def http_post(url = nil, body = nil, headers = nil)
+  def raw_post(url = nil, body = nil, headers = nil)
     block = Proc.new if block_given?
     http_method :post, url, body, headers, &block
   end
--- a/lib/cc/services/asana.rb
+++ b/lib/cc/services/asana.rb
@@ -51,7 +51,7 @@ private
     params = generate_params(name)
     authenticate_http
     http.headers["Content-Type"] = "application/json"
-    post(ENDPOINT, params.to_json) do |response|
+    service_post(ENDPOINT, params.to_json) do |response|
       body = JSON.parse(response.body)
       id = body['data']['id']
       url = "https://app.asana.com/0/#{config.workspace_id}/#{id}"
--- a/lib/cc/services/asana.rb
+++ b/lib/cc/services/asana.rb
@@ -25,6 +25,10 @@ class CC::Service::Asana < CC::Service
     result.merge(
       message: "Ticket <a href='#{result[:url]}'>#{result[:id]}</a> created."
     )
+  rescue CC::Service::HTTPError => ex
+    body = JSON.parse(ex.response_body)
+    ex.user_message = body["errors"].map{|e| e["message"] }.join(" ")
+    raise ex
   end
 
 
--- a/lib/cc/services/asana.rb
+++ b/lib/cc/services/asana.rb
@@ -14,20 +14,19 @@ class CC::Service::Asana < CC::Service
     validates :workspace_id, presence: true
   end
 
+  ENDPOINT = "https://app.asana.com/api/1.0/tasks"
+
   self.title = "Asana"
   self.description = "Create tasks in Asana"
   self.issue_tracker = true
 
   def receive_test
     result = create_task("Test task from Code Climate")
-
-    {
-      ok: true,
-      message: "Ticked <a href='#{result[:url]}'>#{result[:id]}</a> created."
-    }
-  rescue => ex
-    { ok: false, message: ex.message }
+    result.merge(
+      message: "Ticket <a href='#{result[:url]}'>#{result[:id]}</a> created."
+    )
   end
+
 
   def receive_quality
     title = "Refactor #{constant_name} from #{rating} on Code Climate"
@@ -45,6 +44,18 @@ class CC::Service::Asana < CC::Service
 private
 
   def create_task(name)
+    params = generate_params(name)
+    authenticate_http
+    http.headers["Content-Type"] = "application/json"
+    post(ENDPOINT, params.to_json) do |response|
+      body = JSON.parse(response.body)
+      id = body['data']['id']
+      url = "https://app.asana.com/0/#{config.workspace_id}/#{id}"
+      { id: id, url: url }
+    end
+  end
+
+  def generate_params(name)
     params = {
       data: { workspace: config.workspace_id, name: name }
     }
@@ -58,18 +69,11 @@ private
       params[:data][:assignee] = config.assignee
     end
 
-    http.headers["Content-Type"] = "application/json"
+    params
+  end
+
+  def authenticate_http
     http.basic_auth(config.api_key, "")
-
-    url = "https://app.asana.com/api/1.0/tasks"
-    res = http_post(url, params.to_json)
-
-    body = JSON.parse(res.body)
-
-    id = body['data']['id']
-    url = "https://app.asana.com/0/#{config.workspace_id}/#{id}"
-
-    { id: id, url: url }
   end
 
 end
--- a/lib/cc/services/flowdock.rb
+++ b/lib/cc/services/flowdock.rb
@@ -54,7 +54,6 @@ class CC::Service::Flowdock < CC::Service
     }
 
     url = "#{BASE_URL}/messages/team_inbox/#{config.api_token}"
-    http.headers['Content-Type']  = 'application/json'
     http.headers["User-Agent"] = "Code Climate"
 
     post(url, params)
--- a/lib/cc/services/flowdock.rb
+++ b/lib/cc/services/flowdock.rb
@@ -56,6 +56,6 @@ class CC::Service::Flowdock < CC::Service
     url = "#{BASE_URL}/messages/team_inbox/#{config.api_token}"
     http.headers["User-Agent"] = "Code Climate"
 
-    post(url, params)
+    service_post(url, params)
   end
 end
--- a/lib/cc/services/flowdock.rb
+++ b/lib/cc/services/flowdock.rb
@@ -13,11 +13,9 @@ class CC::Service::Flowdock < CC::Service
   self.description = "Send messages to a Flowdock inbox"
 
   def receive_test
-    notify("Test", repo_name, formatter.format_test)
-
-    { ok: true, message: "Test message sent" }
-  rescue => ex
-    { ok: false, message: ex.message }
+    notify("Test", repo_name, formatter.format_test).merge(
+      message: "Test message sent"
+    )
   end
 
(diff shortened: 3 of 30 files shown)


calavera:
It looks good. I guess you've already seen how to extend services to store data in codeclimate, but just in case, check the pull request service extension out: https://github.com/codeclimate/codeclimate/blob/master/lib/cc/services/extensions/github_pull_requests.rb

jyurek:
I have seen that, but I think these things should happen in this gem and not in an extension. Right?

calavera:
Right, the response still needs to be handled in the services gem.

lib/cc/service/http.rb
```diff
@@ -15,6 +15,34 @@ def default_http_options
     end
   end
 
+  def post(url, body = nil, headers = nil, &block)
+    begin
```
calavera: this `begin` is not necessary, either it is the `end` that goes with it.

lib/cc/service/http.rb
```diff
@@ -15,6 +15,34 @@ def default_http_options
...
+      block ||= lambda{|*args| Hash.new }
+      response = http_post(url, body, headers)
+      {
+        ok: response.success?,
+        params: body.as_json,
+        endpoint_url: url,
+        http_status: response.status
+      }.merge(block.call(response))
+    rescue CC::Service::HTTPError => e
```
calavera: these two `rescue` blocks can be one if you set `http_status` like this:

```
http_status: e.respond_to?(:status) ? e.status : nil
```

lib/cc/services/asana.rb
```diff
@@ -45,6 +40,18 @@ def receive_vulnerability
 private
 
   def create_task(name)
+    params = generate_params(name)
+    authenticate_http
+    http.headers["Content-Type"] = "application/json"
+      post(ENDPOINT, params.to_json) do |result|
```
calavera: there is an extra indentation level here.

lib/cc/services/github_pull_requests.rb
```diff
@@ -87,21 +86,24 @@ def add_comment
...
-    http_post(base_status_url("0" * 40), "{}")
-
-  rescue HTTPError => ex
-    if ex.status == 422 # response message: "No commit found for SHA"
-      { ok: true, message: "OAuth token is valid" }
-    else ex.status == 401 # response message: "Bad credentials"
-      { ok: false, message: ex.message }
+    url = base_status_url("0" * 40)
+    params = {}
+    begin
```
calavera: this `begin` is not necessary.

lib/cc/service/http.rb
jyurek: That's true, but I think separating the exceptions is more appropriate and clearer. That the return values happen to be similar is not a reason to introduce a less clear but of code, IMHO.

lib/cc/service/http.rb
```diff
@@ -15,7 +15,37 @@ def default_http_options
...
+    }.merge(block.call(response))
+  rescue CC::Service::HTTPError => e
+    {
+      ok: false,
+      params: body.as_json,
+      http_status: e.status,
+      endpoint_url: url,
+      exception_message: "#{e.class}: #{e.message}"
+    }
+  rescue Exception => e
```
JacobNinja: This should probably only rescue `StandardError` type exceptions

jyurek: Damnit, you're right. I'm usually the one catching this on others' code, too.

jyurek:
The way this works is by moving all of the actual HTTP POSTing into a dedicated method that wraps the response and also handles failures. A successful request will be parsed for standard things like status code, the params we sent, the url we hit, and whether it worked or not. An HTTP error will also include the exception message. Any other error will forego the HTTP code. This is all tested and common to each service, so they don't need their own specific tests for it.

A successful request can be modified by adding extra attributes into the hash via a block. The response body is passed to the block (since we get both JSON and XML from various services). Relevant values can be extracted in this block. Parse errors will trigger the generic error handling and return a "this failed" message to the caller. The uniquely-added attributes are tested on a per-service basis.

Some services make multiple requests. Only the last one is cared about.

lib/cc/service/http.rb
```diff
@@ -15,7 +15,41 @@ def default_http_options
...
+    }.merge(block.call(response))
+  rescue CC::Service::HTTPError => e
+    {
+      ok: false,
+      params: body.as_json,
+      http_status: e.status,
+      endpoint_url: url,
+      exception_message: "#{e.class}: #{e.message}"
+    }
+  rescue *RERAISE_ERRORS
```
jeffrafter: This pattern is making me a little quesy-- can we remove this special casing for test errors and make the following rescue for `StandardError` more specific so it doesn't catch test assertion errors?

jyurek:
The places where `simple_failure` exists in this change, we simply weren't doing anything. The case didn't exist, and so we returned `nil` -- no bugsnag. Since we want to return `ok: true/false` in all cases, now, the case needed to be filled in.

lib/cc/services/asana.rb
```diff
@@ -14,21 +14,24 @@ class Config < CC::Service::Config
...
-    }
-  rescue => ex
-    { ok: false, message: ex.message }
+    result.merge(
+      message: "Ticket <a href='#{result[:url]}'>#{result[:id]}</a> created."
+    )
+  rescue CC::Service::HTTPError => ex
+    body = JSON.parse(ex.response_body)
+    ex.user_message = body["errors"].map{|e| e["message"] }.join(" ")
+    raise ex
```
noahd1: A bit confusing to me - where does this exception get handled? Wonder if the `ErrorHandling` middleware is encouraging handling logic through exceptions.

noahd1:
Questions:

* Since all the services are changing in some way, have we / can we mitigate the risk of this release by end to end testing as much as possible? Having good unit tests is necessary but probably not sufficient in this case.
* I'm a bit uncomfortable with how and where our "response" (the hash) is generated -- in `http.rb`, `with_error_handling.rb` and then mangled further by individual services. There'll always be some degree of per-service customization of our responses, but wondering if we can minimize the distinction between a successful and unsuccessful response -- aka, is there a world where we can generate the success/failure responses in the same class, maybe by making `with_error_handling.rb` more generic - like a `response_generator.rb`? 
* Related to above, failure cases being handled with exceptions (`WithErrorHandling`) right now, but they aren't super exceptional: if a user commits a bad "token" for a service, we expect an error back from the service.

/cc @jeffrafter

lib/cc/services/asana.rb
jyurek: It's no so much logic as it is that we need to augment the exception. There's information that the `receive_test` methods need to put into the replies that can't be handled any other way. That's why we're re-raising the error -- the `WithErrorHandling` middleware handles the errors just fine, but the since the `receive_test` handlers are user-facing, they want/need to put pretty messages on the data returned.

It's possible we also want to handle this in the web-side, though that will be a bit more complicated.

jyurek:
* I'm not sure what more is expected of these tests. Do you want them to actually hit the network? That code isn't really being touched. Only the replies are cared about. I'm all for integration testing, but without hitting the network or using something like VCR we're unable to get any more visibility.
* To be honest, I'm not in love with it either, but I'm not entirely sure how we could have the `WithErrorHandling` class access data that a successful HTTP post gets from the server. Even if we wanted to do such a thing, I think it would require odd layer-crossings and make the code be less understandable. We could prevent mangling the hash by the `receive_test` handlers by letting the thing calling the handler mangle the result, but that would need to either be super general (and therefore less useful) or make the code a nest of `if` statements.
* We've chosen to treat any non-2XX status code as an error to be raised. We can rescue certain errors in the handlers themselves and build responses for things like 401 and 422, but this will be overshadowed by those cases (see `GitHubPullRequestService`) where we need to see those in the raw. While I believe there is a better way to handle things, I feel as though they're outside the scope of this change, which was to give the services useful responses when their event handlers are hit.

noahd1:
We can discuss these more tomorrow.

* Yeah, wasn't commenting on the automated test suite which as far as I can tell is quite good. I more meant that in cases like this we still need to be careful and probably do manual QA and/or look at other strategies to minimize risk.
* Sorry, I misspoke, I didn't mean `WithErrorHandling` which I agree isn't really a place for this kind of generic behavior -- I meant producing an kind of response object in a Faraday middleware, aka, replacing `ResponseCheck`. Downside would be you would probably have to stick the resulting object back in the `env` to have it be available later
* I agree there's some changes that are outside the scope of this PR. Just trying to find a balance here. Easier to chat about in person.

lib/cc/service/http.rb
```diff
@@ -15,7 +15,23 @@ def default_http_options
...
+  def post(url, body = nil, headers = nil, &block)
+    block ||= lambda{|*args| Hash.new }
+    response = raw_post(url, body, headers)
+    {
+      ok: response.success?,
+      params: body.as_json,
+      endpoint_url: url,
+      status: response.status,
+      message: "Success"
+    }.merge(block.call(response))
```
noahd1: After looking at the code more, I now see that `post` returns a hash but `get` returns, as it did before, some kind of http response object. I think this mismatch in the API is confusing, which is why I'm trying to explore if there's a better place for "response generation" that exists outside of this class. See main thread.

jyurek: Yes, a GET should generate the same kind of response.

lib/cc/services/flowdock.rb
```diff
@@ -13,11 +13,9 @@ class Config < CC::Service::Config
...
 
   def receive_test
-    notify("Test", repo_name, formatter.format_test)
-
-    { ok: true, message: "Test message sent" }
-  rescue => ex
-    { ok: false, message: ex.message }
+    notify("Test", repo_name, formatter.format_test).merge(
+      message: "Test message sent"
+    )
```
jeffrafter: Is the rescue lost here?

lib/cc/services/flowdock.rb
```diff
@@ -56,7 +54,9 @@ def notify(subject, project, content)
     }
 
     url = "#{BASE_URL}/messages/team_inbox/#{config.api_token}"
+    http.headers['Content-Type']  = 'application/json'
```
jeffrafter: This is a material change... it may be important can we separate it out though?

lib/cc/services/github_pull_requests.rb
```diff
@@ -59,31 +43,56 @@ def receive_pull_request
...
+      update_status_error
+    else
+      simple_failure("Unknown state")
     end
   end
 
 private
 
+  def simple_failure(message)
+    { ok: false, message: message }
```
jeffrafter: Is this the correct hash for failure?

lib/cc/services/github_pull_requests.rb
```diff
@@ -93,37 +102,44 @@ def add_comment
...
     end
   end
 
   def receive_test_status
-    http_post(base_status_url("0" * 40), "{}")
-
-  rescue HTTPError => ex
-    if ex.status == 422 # response message: "No commit found for SHA"
-      { ok: true, message: "OAuth token is valid" }
-    else ex.status == 401 # response message: "Bad credentials"
```
jeffrafter: Missing the 401 rescue, this error handling is probably useful for debugging bad credentials

lib/cc/services/github_pull_requests.rb
```diff
@@ -93,37 +102,44 @@ def add_comment
...
+    if e.status == 422
+      {
+        ok: true,
+        params: params.as_json,
+        status: e.status,
+        endpoint_url: url,
+        message: "OAuth token is valid"
+      }
+    else
+      raise
```
jeffrafter: Does this raise yield the correct status/hash

jeffrafter: See the rescue that was removed below (we used to handle all exceptions out of here)

lib/cc/services/hipchat.rb
```diff
@@ -18,11 +18,9 @@ class Config < CC::Service::Config
   self.description = "Send messages to a HipChat chat room"
 
   def receive_test
-    speak(formatter.format_test, "green")
-
-    { ok: true, message: "Test message sent" }
-  rescue => ex
```
jeffrafter: Is this rescue handled up the callstack? I think so... just confirming

lib/cc/services/flowdock.rb
jyurek: Not really, no. The errors weren't being handled before, but they are now -- by the `WithErrorHandling` middleware. It's not in this method, but the previous version never did anything with the error anyway.

lib/cc/services/flowdock.rb
jyurek: Yeah this doesn't need to be here.

lib/cc/services/github_pull_requests.rb
jyurek: Yup. Pretty much all failures of this type were of this format.

lib/cc/services/github_pull_requests.rb
jyurek: The 401s will get caught by the `WithErrorHandling` middleware and doesn't need to be a special case here anymore.

lib/cc/services/github_pull_requests.rb
jyurek: Yes, it will. We only care about 422 here, as that's a "success". Anything else will be, as mentioned above, handled by the `WithErrorHandling` middleware.

lib/cc/services/hipchat.rb
jyurek: Yup! See above. :)

PR #55, opened 2015-03-16
Update GitHub status description for skipped builds

mrgilman:
https://trello.com/c/q3ByzTqY

When analysis for a commit is skipped because a newer commit was pushed
to the branch, we want to set the status to "success" so that it isn't
stuck in a yellow "pending" state, but we also want to indicate that
analysis didn't run on the skipped commit. So we differentiate the
skipped case with a different status message. Also, never add a comment
for a commit that we have skipped, regardless of the config settings.

### DIFF
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -55,6 +55,10 @@ class CC::Service::GitHubPullRequests < CC::Service
       else
         simple_failure("Nothing happened")
       end
+    when "skipped"
+      if config.update_status
+        update_status_skipped
+      end
     when "error"
       update_status_error
     else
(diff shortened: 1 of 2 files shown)


lib/cc/services/github_pull_requests.rb
```diff
@@ -45,16 +45,14 @@ def receive_pull_request
...
       update_status_pending
     when "success"
-      if config.update_status && config.add_comment
+      if config.update_status || config.add_comment
         update_status_success
         add_comment
-      elsif config.update_status
-        update_status_success
-      elsif config.add_comment
-        add_comment
```
marshally: My spidey sense is tingling about this change around `config.add_comment`.

Are there any conditions where we want to call `add_comment` and not call `update_status_success` -- /cc @noahd1 @jonathancadepowers

marshally:
:shipit:

PR #56, opened 2015-03-16
Give an OK of false if the PR comment exists

jyurek:


lib/cc/services/github_pull_requests.rb
```diff
@@ -85,19 +85,17 @@ def update_status_pending
...
+      state:       state,
+      description: description,
+      target_url:  @payload["details_url"],
+      context:     "codeclimate"
+    }
+    service_post(status_url, params.to_json)
   end
 
   def add_comment
-    if config.add_comment && !comment_present?
```
marshally: Why was the `add_comment` check removed?

iancanderson: We are actually relying on these checks to be here (and in `#update_status`) in our new PR :smile: https://github.com/codeclimate/codeclimate-services/pull/55

iancanderson: Okay - we resolved our differences and are no longer relying on the config check that @jyurek removed here - so #55 should be compatible with this PR now

jyurek: @marshally Because the check actually happens here: https://github.com/codeclimate/codeclimate-services/blob/jy-do-not-allow-nil-branches/lib/cc/services/github_pull_requests.rb#L53 and this check only served to muddy the waters about what was happening.

jyurek:
Talked with @marshally on the phone. He gives a :shipit: by proxy.

PR #57, opened 2015-03-18
Do not send status update if user disable it

sikachu:
We forgot to check if user has turned on the `update_status` flag or not before we sends a pending status to GitHub.

This commit includes refactoring, hopefully to cleanup the confusion on conditionals.

### DIFF
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -40,30 +40,15 @@ class CC::Service::GitHubPullRequests < CC::Service
 
   def receive_pull_request
     setup_http
+    state = @payload["state"]
 
-    case @payload["state"]
-    when "pending"
-      update_status_pending
-    when "success"
-      if config.update_status && config.add_comment
-        update_status_success
-        add_comment
-      elsif config.update_status
-        update_status_success
-      elsif config.add_comment
-        add_comment
-      else
-        simple_failure("Nothing happened")
-      end
-    when "skipped"
-      if config.update_status
-        update_status_skipped
-      end
-    when "error"
-      update_status_error
+    if %w(pending success skipped error).include?(state)
+      send("update_status_#{state}")
     else
-      simple_failure("Unknown state")
+      @response = simple_failure("Unknown state")
     end
+
+    response
   end
 
 private
@@ -72,19 +57,27 @@ private
     { ok: false, message: message }
   end
 
+  def response
+    @response || simple_failure("Nothing happened")
+  end
+
   def update_status_skipped
-    update_status("success", "Code Climate has skipped analysis of this commit.")
+    update_status(
+      "success",
+      "Code Climate has skipped analysis of this commit."
+    )
   end
 
   def update_status_success
     update_status("success", "Code Climate has analyzed this pull request.")
+    add_comment
   end
 
   def update_status_error
     update_status(
       "error",
       "Code Climate encountered an error while attempting to analyze this " +
-      "pull request."
+        "pull request."
     )
   end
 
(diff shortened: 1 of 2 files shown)


test/github_pull_requests_test.rb
```diff
@@ -63,6 +63,27 @@ def test_no_status_update_for_skips_when_update_status_config_is_falsey
...
+    })
+  end
+
+  def test_no_status_update_for_error_when_update_status_config_is_falsey
+    # With no POST expectation, test will fail if request is made.
+
+    receive_pull_request({}, {
+      github_slug: "pbrisbin/foo",
+      commit_sha:  "abc123",
+      state:       "pending",
```
iancanderson: this should be `state: "error"` to match the test description

lib/cc/services/github_pull_requests.rb
```diff
@@ -40,30 +40,34 @@ def receive_test
 
   def receive_pull_request
```
noahd1: This method is getting pretty long and hard to read. Ideas on breaking it down? De-compose the method?

test/github_pull_requests_test.rb
sikachu: I fixed it but you commented the wrong one.

sikachu:
This is ready for another look. I updated the code to hopefully make it cleaner, and also add more tests on those default response ... Ready for re-review!

lib/cc/services/github_pull_requests.rb
```diff
@@ -40,30 +40,15 @@ def receive_test
...
-        add_comment
-      else
-        simple_failure("Nothing happened")
-      end
-    when "skipped"
-      if config.update_status
-        update_status_skipped
-      end
-    when "error"
-      update_status_error
```
marshally: Goodbye cruel conditionals!

lib/cc/services/github_pull_requests.rb
marshally: @noahd1 - I think the new `#receive_pull_request` is easier to understand. wdyt?

iancanderson:
LGTM

lib/cc/services/github_pull_requests.rb
noahd1: :+1:

noahd1:
LGTM

lib/cc/services/github_pull_requests.rb
```diff
@@ -72,19 +57,27 @@ def simple_failure(message)
...
-    update_status("success", "Code Climate has skipped analysis of this commit.")
+    update_status(
+      "success",
+      "Code Climate has skipped analysis of this commit."
+    )
   end
 
   def update_status_success
     update_status("success", "Code Climate has analyzed this pull request.")
+    add_comment
```
noahd1: Not in love with having status mixed in with commenting, but it's a larger issue with this class in general (that it does two things). Commenting should go away soon, so I'm ok with this. 

However -- I think we swap the order of these lines for now. We're not capturing both results in the response back (just the last one), but the PR one is the more important one right now.

sikachu: Right. This class is doing too much.

I'm fine with the swap, but that means we're actually change the functionality of this (and its return value). Currently, [we return the status of add comment](https://github.com/codeclimate/codeclimate-services/blob/4aa7dfaa1a09438e41c27afa4bcebbe8ea3c9e58/lib/cc/services/github_pull_requests.rb#L49-L50) only. Do you think we should swap it anyway?

lib/cc/services/github_pull_requests.rb
```diff
@@ -93,27 +86,34 @@ def update_status_pending
...
+          body: COMMENT_BODY % @payload["compare_url"]
+        }.to_json
+
+        @response = service_post(comments_url, body) do |response|
+          doc = JSON.parse(response.body)
+          { id: doc["id"] }
+        end
+      else
+        @response = {
+          ok: false,
```
noahd1: This may be a pre-existing problem but I don't think this branch should respond with `ok: false` -- it should be `ok: true`.

My understanding is that an `ok: false` has user facing implications: it's a red stop sign meaning the integration isn't working properly. However, the comment existing on a PR if we've already commented is _expected_. It happens if you open a PR and then push more commits (so a lot).

lib/cc/services/github_pull_requests.rb
noahd1: That's fair. Perhaps a separate PR then is appropriate for this change. I think it's in scope however, because we're now capturing the result of these calls -- and the most important result is the result of posting a status to a PR (not the comment).

sikachu: Yep. Let's do that swap in another PR. :egg:

lib/cc/services/github_pull_requests.rb
sikachu: That's also the [current behavior](https://github.com/codeclimate/codeclimate-services/blob/4aa7dfaa1a09438e41c27afa4bcebbe8ea3c9e58/lib/cc/services/github_pull_requests.rb#L116). If we're to change it, let's do it in another PR/card.

But yeah, I agree with you. It used to be the case where the return value may/may not matter, but now it does matter.

PR #62, opened 2015-04-07
Show new PR status message only for repos with beta feature enabled

iancanderson:
Also - use new nested hash structure to get issue counts from the event payload.

https://trello.com/c/f17uUGRd

### DIFF
--- a/lib/cc/service.rb
+++ b/lib/cc/service.rb
@@ -69,10 +69,11 @@ module CC
       end
     end
 
-    def initialize(config, payload)
-      @payload = payload.stringify_keys
-      @config  = create_config(config)
-      @event   = @payload["name"].to_s
+    def initialize(config, payload, repo_config)
+      @payload     = payload.stringify_keys
+      @config      = create_config(config)
+      @event       = @payload["name"].to_s
+      @repo_config = repo_config
 
       load_helper
       validate_event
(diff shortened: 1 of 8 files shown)


iancanderson:
@codeclimate/review

lib/cc/service.rb
```diff
@@ -69,10 +69,11 @@ def self.slug
       end
     end
 
-    def initialize(config, payload)
-      @payload = payload.stringify_keys
-      @config  = create_config(config)
-      @event   = @payload["name"].to_s
+    def initialize(config, payload, beta_service_features)
```
noahd1: Hmm. Not sure I like the name so much. How about something more along the lines of `feature_toggle` ? Whether its beta or not is sort of a separate concern.

noahd1: seems like the toggler could be an optional parameter

iancanderson: I agree with the name change. I don't think it needs to be optional, because the app only instantiates this object in one place)

test/github_pull_requests_presenter_test.rb
```diff
@@ -1,51 +1,58 @@
...
 
 private
 
-  def message_from_payload(payload)
-    CC::Service::GitHubPullRequests::Presenter.new(payload).success_message
+  def build_payload(issue_counts)
+    { "issue_comparison_counts" => issue_counts }
+  end
+
+  def build_presenter(beta_enabled, issue_counts)
```
noahd1: `beta_enabled` reference

iancanderson: ah, thanks

noahd1:
LGTM other than reference to "beta"

sikachu:
:+1:

PR #63, opened 2015-04-07
Add require to github pull requests presenter

iancanderson:
Since the presenter is an inner class of the service class


iancanderson:
Closing this in favor of an upcoming better solution

PR #64, opened 2015-04-08
Fix load-order dependency

iancanderson:
This fixes an `unitialized constant CC::Service::GitHubPullRequests`,
which was happening if `CC::Service::GithubPullRequests::Presenter` was loaded
before `CC::Service::GithubPullRequests`. Loading order is non-deterministic and
can differ depending on OS, since services are loaded via `Dir[]`).

Going forward, services that need a presenter class can require the presenter
themselves.


lib/cc/presenters/github_pull_requests_presenter.rb
```diff
@@ -0,0 +1,62 @@
+module CC
```
iancanderson: This file was just moved from `lib/cc/services` to a new `lib/cc/presenters` directory, and nested under `module CC` and `class Service`, which less susceptible to load order issues than doing `class CC::Service::GithubPullRequestsPresenter`

noahd1:
LGTM

PR #66, opened 2015-04-08
Sort paths before requiring them

iancanderson:
Since the order of files returned from `Dir[]` is non-deterministic and depends
on OS. See http://ruby-doc.org/core-2.1.2/Dir.html#method-c-glob


noahd1:
:+1:

PR #67, opened 2015-04-09
Return nil when a service doesn't respond to an event

mrgilman:
Returning `ok: false` sets `working` to `false` on the service document
in the app, which is incorrect when the service isn't being invoked
because it isn't defined for the given event type. Returning `nil` makes
it easier to distinguish between services that are actually not working
and services that are a no-op for a given event.


noahd1:
:+1:

PR #68, opened 2015-04-14
Flag ignored service invocations

mrgilman:
https://trello.com/c/1QAKfYtf

Distinguish between service invocations that fail and service
invocations that don't run because the service doesn't respond to an
event without returning nil, since nil is an unexpected response in the
middleware.


noahd1:
:+1:

PR #69, opened 2015-04-16
Return hash instead of nil when no changes in snapshot

mrgilman:
https://trello.com/c/FVI4JXBI

Before this change, attempting to send a snapshot to Slack when there
were no alert or improved constants would return nil, which is an
unexpected value in the app. This change checks to see if those changes
exist before attempting to send a snapshot, and if they don't, returns
a hash with ignored: true instead of nil.

### DIFF
--- a/lib/cc/formatters/snapshot_formatter.rb
+++ b/lib/cc/formatters/snapshot_formatter.rb
@@ -59,6 +59,10 @@ module CC::Formatters
         @improved_constants_payload = data.merge("constants" => improved_constants) if improved_constants.any?
       end
 
+      def changed?
+        alert_constants_payload || improved_constants_payload
+      end
+
     private
 
       def new_constants_selector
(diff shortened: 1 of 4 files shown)


lib/cc/formatters/snapshot_formatter.rb
```diff
@@ -59,6 +59,10 @@ def initialize(payload)
         @improved_constants_payload = data.merge("constants" => improved_constants) if improved_constants.any?
       end
 
+      def changed?
+        alert_constants_payload || improved_constants_payload
```
GordonDiggs: What do you think of wrapping this line in `!!()` so that the method actually returns a boolean value (as suggested by its signature) rather than either payload?

mrgilman: I'm not a fan of `!!` myself, but I agree about returning a boolean. Updated.

GordonDiggs:
LGTM! :shipit:

lib/cc/formatters/snapshot_formatter.rb
GordonDiggs: I actually like your solution better!

mrgilman: Yay ActiveSupport!

PR #70, opened 2015-04-20
Remove rollout call for quality stats

GordonDiggs:
Now that this feature is enabled for all repos, we can remove the check

### DIFF
--- a/lib/cc/presenters/github_pull_requests_presenter.rb
+++ b/lib/cc/presenters/github_pull_requests_presenter.rb
@@ -13,14 +13,10 @@ module CC
       end
 
       def success_message
-        if issue_counts_in_payload?
-          if both_issue_counts_zero?
-            "Code Climate didn't find any new or fixed issues."
-          else
-            "Code Climate found #{formatted_issue_counts}."
-          end
+        if both_issue_counts_zero?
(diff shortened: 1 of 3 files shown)


lib/cc/presenters/github_pull_requests_presenter.rb
```diff
@@ -13,7 +13,7 @@ def initialize(payload, repo_config)
       end
 
       def success_message
-        if @repo_config.pr_status_quality_stats?
+        if issue_counts_in_payload?
```
noahd1: Is this conditional still needed?

GordonDiggs: The way these issues are passed and particularly how they are set in this class means they could be nil. I figured it was worth it to keep this check so that if the counts aren't in the request, we can still report statuses on github PRs.

noahd1: That might be vestigial. Are you aware of any cases where it makes sense that they'd be `nil`?

sikachu: Yeah. If we are the one controlling the message, I think this conditional is unnecessary. YAGNI.

noahd1:
:+1:

PR #75, opened 2015-06-16
All delimiters to numbers in PR status

brynary:


pbrisbin:
Commit message is a typo (only mentioning b/c this is OSS). LGTM

PR #76, opened 2015-07-01
update failed pr message

ABaldwinHunter:
@codeclimate/review WIP. I wanted to create a more customized message for PR's that error due to incompatible worker versions- particularly engine related.

So, my solution was:

1) add a key to payload that specifies if the worker version was changed (also updated in finalizer)
PR: https://github.com/codeclimate/finalizer/pull/49
2) create two constants for error messages: one default, one specific to worker version issue.

EDIT: Feedback welcome. It's my first time working with `codeclimate-services`. If my approach seems right, I'll add tests.

### DIFF
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -21,7 +21,6 @@ class CC::Service::GitHubPullRequests < CC::Service
   BASE_URL = "https://api.github.com"
   BODY_REGEX = %r{<b>Code Climate</b> has <a href=".*">analyzed this pull request</a>}
   COMMENT_BODY = '<img src="https://codeclimate.com/favicon.png" width="20" height="20" />&nbsp;<b>Code Climate</b> has <a href="%s">analyzed this pull request</a>.'
-
   # Just make sure we can access GH using the configured token. Without
   # additional information (github-slug, PR number, etc) we can't test much
   # else.
@@ -87,8 +86,7 @@ private
   def update_status_error
     update_status(
       "error",
-      "Code Climate encountered an error while attempting to analyze this " +
-        "pull request."
+      message
     )
   end
 
@@ -192,6 +190,10 @@ private
 
   def commit_sha
     @payload.fetch("commit_sha")
+  end
+
+  def message
+    @payload.fetch("message")
   end
 
   def number
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -24,6 +24,11 @@ class CC::Service::GitHubPullRequests < CC::Service
   # Just make sure we can access GH using the configured token. Without
   # additional information (github-slug, PR number, etc) we can't test much
   # else.
+
+  MESSAGES = [
+    DEFAULT_ERROR = "Code Climate encountered an error attempting to analyze this pull request",
+  ]
+
   def receive_test
     setup_http
 
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -91,7 +91,7 @@ private
   def update_status_error
     update_status(
       "error",
-      message
+      @payload["message"] || DEFAULT_ERROR
     )
   end
 
@@ -195,10 +195,6 @@ private
 
   def commit_sha
     @payload.fetch("commit_sha")
-  end
-
-  def message
-    @payload.fetch("message")
   end
 
   def number
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -107,6 +107,9 @@ private
         target_url:  @payload["details_url"],
         context:     "codeclimate"
       }
+      if state == "error"
+        params.delete(:target_url)
+      end
       @response = service_post(status_url, params.to_json)
     end
   end
--- a/test/github_pull_requests_test.rb
+++ b/test/github_pull_requests_test.rb
@@ -69,6 +69,7 @@ class TestGitHubPullRequests < CC::Service::TestCase
       github_slug: "pbrisbin/foo",
       commit_sha:  "abc123",
       state:       "error",
+      message:     "Code Climate encountered an error while attempting to analyze this pull request.",
     })
   end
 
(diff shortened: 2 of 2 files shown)


ABaldwinHunter:
@codeclimate/review I thought this PR was ready re-review - but now am not sure why Travis Build is failing (exploring). Tests pass locally when I run `bundle exec rake`.

It updates `pull_requests` to expect a payload with a `[:message]` key that is only used if the `state` is an error.

Goal is to provide clearer feedback to users when PR fails due to incompatible worker versions. 

First draft message in that case: `"Whoops. Code Climate cannot analyze a pull request against its base when only one is configured to use engines. We apologize for the inconvenience."`

See related changes in `finalizer`: https://github.com/codeclimate/finalizer/pull/50

lib/cc/services/github_pull_requests.rb
```diff
@@ -87,8 +86,7 @@ def presenter
   def update_status_error
     update_status(
       "error",
-      "Code Climate encountered an error while attempting to analyze this " +
-        "pull request."
+      message
```
pbrisbin: What do you think about `@payload["message"] || DEFAULT_ERROR_MESSAGE`?

I think generally, all the `update_status_x` methods should use `@payload["message"] || SOME_DEFAULT` and let the callers be in charge of it (not the presenter here), but that's a bigger change and we can punt on it for now.

ABaldwinHunter: @pbrisbin okay that's fine by me. I see some reasons supporting both places as the site of setting a `message_default` - maybe it makes sense to have messages get set in one place, so `finalizer` should worry about the messages and `services` can just eat them. But on the other hand, I think having a default message on `services` side is fine. 

Generally, agree that it seems a little confusing that the presenter is used for some messages and not others (and the method is named `success_message` even when the status is failure :p ). But punting sounds ideal to me. :) Keep changes separate.

pbrisbin: Agreed. cc-services should not be in the business of crafting messages based on data, etc. But having a sane, generic default seems totally reasonable. And is important for something like this which is a shared, open source gem that has to think about backwards compatibility.

ABaldwinHunter: Mmmm makes sense. 

Imagine we'll probably also want to avoid fetching the `message` then.

pbrisbin: Yup, my example snippet should work and make the new `#message` method unused (or keep it but use `#[]` instead).

ABaldwinHunter:
@codeclimate/review I added the changes discussed above, including adding a `DEFAULT_ERROR` message constant.

New change that might need improvement: my naive attempt to remove the `Details` link was to remove the details url (`target_url`) from getting passed as params when the status was `"error"`. Wasn't sure if that was the correct approach. Was also wondering best way to go about testing this sort of change.

GordonDiggs:
:+1: LGTM

lib/cc/services/github_pull_requests.rb
```diff
@@ -104,6 +107,9 @@ def update_status(state, description)
         target_url:  @payload["details_url"],
         context:     "codeclimate"
       }
+      if state == "error"
+        params.delete(:target_url)
```
brynary: At first glance, I don't love this. Why not just not include the target_url in the original payload. I don't think this class should be responsible for "overriding" the payload in this way. /c @gordondiggs @abaldwinhubter @pbrisbin

brynary: Typo. /cc @abaldwinhunter

ABaldwinHunter: @brynary apologies, just saw your note! I believe the thought was that removing the `details_url` would remove the `details` link from PRs that error out.

What's constructed in this `update_status` method doesn't match the original payload perfectly, and without adding in a conditional, it might set the `target_url` to `nil` when `@payload["details_url"]` gets left out.

That said, I think this is the place in `finalizer` where `payload` gets created and sent:
```ruby
  module Finalizer
    class UpdatePullRequest
      def initialize(ref_point)
        @ref_point = ref_point
        @repo = ref_point.repo
        @pull_request = ref_point.pull_request
      end

      def update
        return unless updatable?

        emitter = EventEmitter.new
        emitter.emit(
          "pull_request",
          state: evaluation.state,
          issue_comparison_counts: {
            new: evaluation.new_issue_count,
            fixed: evaluation.fixed_issue_count,
          },
          details_url: evaluation.url,
          compare_url: evaluation.url,
          commit_sha: ref_point.commit_sha,
          github_slug: repo.github_slug,
          repo_id: repo.id.to_s,
          number: pull_request.number,
          message: evaluation.message,
        )
      end
``` 
We could perhaps make a change there.

brynary: Yeah, the goal of removing the link in errored PRs makes sense (temporarily). I was only commenting on the implementation.




It seems like we could either change the payload that is sent to not include the URL (which would require a change elsewhere) or remove it here. What are your thoughts on how it should work?



—Bryan Helmkamp, Founder, Code Climate

bryan@brynary.com / 646-379-1810 / @brynary




Typed with thumbs.

pbrisbin: Agreed that change would be better in finalizer.

> without adding in a conditional, it might set the target_url to nil when @payload["details_url"] gets left out.

I would test if this matters. If it does, I'd rather *that* conditional than the one you have.

ABaldwinHunter: @pbrisbin @brynary Hmm. It seems to me that right now, the `codeclimate-services` entity is not just getting a payload and relaying it, but making some additional calculations - e.g. formatting messages based on number of issues fixed, adding a `context: 'codeclimate'` to params passed to http request. 

However, we recently updated finalizer to also do some analysis, including checking whether errored state is due to worker_version incompatibility and adding a unique message to payload, if that's the case.

On one hand, it seems nice for `finalizer` (or whatever other entity) to just be able to send a bunch of information to `codeclimate-services` github_PR and have cc-services know what to do with it.

On the other hand, it seems nice to have cc-services not know too much about how the stuff it's handling gets calculated - it just knows it needs to send message x to place y. 

So, yeah. Agree. Ultimately I favor minimizing analysis in cc-services, which would mean not having it read `state` to decide whether to add `details_url` or not.

brynary: @ABaldwinHunter OK, can you please address this at your convenience?

PR #77, opened 2015-07-06
bump version

ABaldwinHunter:
@codeclimate/review This PR bumps the Code Climate Services version from 0.1.0 to 0.2.0, after a change to services and finalizer that lets cc-services now receive an optional `"messages"` key on `@payload` - to provide clarity when analysis fails due to version incompatibility between snapshots. I have tested using the PR event trigger script in `codeclimate-services`.

The PR also makes minor updates to the pull request test, bringing over some clarifying instructions re OAUTH token from a same test in app.

If all looks well, I will bump the version and update `codeclimate/app`.

One note: 
Intent of change was to make error messages more explicit and remove details link when comparisons fail due to analysis version incompatibility between snapshots. 

Default error message:
screen shot 2015-07-02 at 2 59 21 pm

Analysis fails due to analysis version incompatibility error message:

screen shot 2015-07-02 at 2 59 49 pm

The lengths of these messages were calculated with the above space in mind. However, when CodeClimate is the only service someone is using (might be rare), the set up for checks is different and less room is available for message, leading to it getting cut off:

screen shot 2015-07-06 at 10 40 07 am

### DIFF
--- a/pull_request_test.rb
+++ b/pull_request_test.rb
@@ -6,6 +6,14 @@
 #
 #   $ OAUTH_TOKEN="..." bundle exec ruby pull_request_test.rb
 #
+#     OAUTH_TOKEN: Personal GitHub access token
+#
+#         GitHub >
+#         Account settings >
+#         Applications >
+#         Personal access tokens >
+#         Generate new token
+#
 ###
 require 'cc/services'
 CC::Service.load_services
@@ -29,6 +37,7 @@ service = CC::Service::GitHubPullRequests.new({
   # https://github.com/codeclimate/nillson/pull/33
   state:       "success",
   github_slug: "codeclimate/nillson",
+  issue_comparison_counts: {"new" => 0, "fixed" => 0},
   number:      33,
   commit_sha:  "986ec903b8420f4e8c8d696d8950f7bd0667ff0c"
 })
--- a/lib/cc/services/version.rb
+++ b/lib/cc/services/version.rb
@@ -1,5 +1,5 @@
 module CC
   module Services
-    VERSION = "0.1.0"
+    VERSION = "0.2.0"
   end
 end


pull_request_test.rb
```diff
@@ -26,11 +47,14 @@ def call
   add_comment:   true,
 }, {
   name:        "pull_request",
-  # https://github.com/codeclimate/nillson/pull/33
-  state:       "success",
-  github_slug: "codeclimate/nillson",
-  number:      33,
-  commit_sha:  "986ec903b8420f4e8c8d696d8950f7bd0667ff0c"
+  # https://github.com/codeclimate/app/pull/1325
```
GordonDiggs: This should use a test repo as an example.

ABaldwinHunter: Thanks! Made switch to test repo.

GordonDiggs:
LGTM

pull_request_test.rb
```diff
@@ -2,9 +2,30 @@
 #
 # Ad-hoc script for updating a pull request using our service.
 #
+###
 # Usage:
 #
-#   $ OAUTH_TOKEN="..." bundle exec ruby pull_request_test.rb
+#   $ script SERVER REPO PR OAUTH_TOKEN SERVICE_TOKEN
```
pbrisbin: This doesn't seem right... is the script really invoked in this new manner?

pbrisbin: Actually are any changes to this script needed?

pbrisbin: To unpack this a little bit:

The script should work as-is with either version of the services gem, so no change should be needed. The idea is it's here with example values that you modify locally to test whatever it is you need to test. If we checked in those modifications every time, the script would see unnecessary churn.

We did identify that the `OAUTH_TOKEN` variable the script comment shows is a little mysterious, so I could see a change to further describe how to find a suitable value for that variable being valuable here.

pbrisbin:
I think you can revert the script changes, but the version bump LGTM.

pull_request_test.rb
ABaldwinHunter: Okay, makes sense. I'll keep a comment about `OAUTH_TOKEN` but revert the rest of script changes.

ABaldwinHunter:
@codeclimate/review Thanks for feedback. Ready for re-review.

I reverted all changes to the script except two. The PR still:

1) adds comment explaining github `OAUTH_TOKEN` source
2) adds key `issue_comparison_counts` to sample payload, which seemed necessary.

Without `issue_comparison_counts`, I got the following error:

```
Ashleys-MacBook-Pro:codeclimate-services ashleybaldwin-hunter$ bundle exec ruby pull_request_test.rb 
warning: ignoring extraneous `ruby-' prefix in version `ruby-2.0.0-p353'
         (set by /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/.ruby-version)
/Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/presenters/github_pull_requests_presenter.rb:26:in `each': undefined method `zero?' for nil:NilClass (NoMethodError)
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/presenters/github_pull_requests_presenter.rb:26:in `all?'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/presenters/github_pull_requests_presenter.rb:26:in `both_issue_counts_zero?'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/presenters/github_pull_requests_presenter.rb:16:in `success_message'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/services/github_pull_requests.rb:79:in `update_status_success'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/services/github_pull_requests.rb:52:in `receive_pull_request'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service.rb:86:in `public_send'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service.rb:86:in `block in receive'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service.rb:84:in `each'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service.rb:84:in `receive'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service/invocation.rb:41:in `block in initialize'
	from pull_request_test.rb:80:in `call'
	from pull_request_test.rb:80:in `call'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service/invocation/invocation_chain.rb:12:in `call'
	from /Users/ashleybaldwin-hunter/Desktop/code_climate/codeclimate-services/lib/cc/service/invocation.rb:45:in `initialize'
	from pull_request_test.rb:97:in `new'
	from pull_request_test.rb:97:in `<main>'
```

`GitHubPullRequestPresenter` expects there to be a key `issue_comparison_count` in payload. Not sure how that was left out before. 

from `github_pull_requests_presenter.rb`:

```ruby
module CC
  class Service
    class GitHubPullRequestsPresenter
      include ActiveSupport::NumberHelper

      def initialize(payload)
        issue_comparison_counts = payload["issue_comparison_counts"]

        if issue_comparison_counts
          @fixed_count = issue_comparison_counts["fixed"]
          @new_count = issue_comparison_counts["new"]
        end
      end
```
and presenter gets created in `github_pull_requests.rb`:
```ruby
  def update_status_success
    add_comment
    update_status("success", presenter.success_message)
  end

  def update_status_failure
    add_comment
    update_status("failure", presenter.success_message)
  end

  def presenter
    CC::Service::GitHubPullRequestsPresenter.new(@payload)
  end
```

pbrisbin:
LGTM

PR #78, opened 2015-07-06
keep comment with target method

ABaldwinHunter:
@codeclimate/review @pbrisbin This PR fixes the placement of a comment that I'd separated from target method.


pbrisbin:
LGTM

PR #79, opened 2015-08-24
Add Issue type

GordonDiggs:
/cc @codeclimate/review 

This adds an `issue` type for messages sent to external services. I chose to add this new type over adding to `quality`, as `quality` is already used for 2 things (rating changes and refactoring hot spots). 

I realize there is a fair amount of duplicated code here - I'm open to ideas on how to fix that, but I'm not familiar enough with this codebase to make that call right now.

I skipped the following services, as they only deal with transitions:

* Campfire
* Flowdock
* Hipchat
* Slack

Preview on each service:


pbrisbin:
LGTM

PR #80, opened 2015-09-09
Add support for specifying a separate Github base URL

GordonDiggs:
When using GH:E, api.github.com will not be the endpoint used to handle
PRs and issues. We can specify it in the configuration to handle that
case better. Defaulting it to api.github.com means that we do not need
to backfill any data.

@codeclimate/review

### DIFF
--- a/lib/cc/services/github_issues.rb
+++ b/lib/cc/services/github_issues.rb
@@ -9,6 +9,10 @@ class CC::Service::GitHubIssues < CC::Service
     attribute :labels, String,
       label: "Labels (comma separated)",
       description: "Comma separated list of labels to apply to the issue"
+    attribute :base_url, String,
+      label: "Github API Base URL",
+      description: "Base URL for the Github API",
+      default: "https://api.github.com"
 
     validates :oauth_token, presence: true
   end
@@ -16,8 +20,6 @@ class CC::Service::GitHubIssues < CC::Service
   self.title = "GitHub Issues"
   self.description = "Open issues on GitHub"
   self.issue_tracker = true
-
-  BASE_URL = "https://api.github.com"
 
   def receive_test
     result = create_issue("Test ticket from Code Climate", "")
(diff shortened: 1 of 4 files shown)


lib/cc/services/github_issues.rb
```diff
@@ -55,6 +55,10 @@ def receive_issue
 
 private
 
+  def base_url
+    ENV["CODECLIMATE_GITHUB_API_URL"] || DEFAULT_BASE_URL
```
pbrisbin: Can this defaulting happen in `CodeClimate::Config` and use `CodeClimate.github_api_url` everywhere?

GordonDiggs: This gem won't always have access to that config object, will it?

GordonDiggs: We do default there, but I didn't think it was re-usable.

test/github_pull_requests_test.rb
```diff
@@ -244,6 +244,20 @@ def test_pull_request_nothing_happened
...
+    ENV["CODECLIMATE_GITHUB_API_URL"] = "http://example.com"
+
+    @stubs.get("/user") do |env|
+      assert env[:url].to_s == "http://example.com/user"
+      [200, { "x-oauth-scopes" => "gist, user, repo" }, ""]
+    end
+
+    assert receive_test({ add_comment: true })[:ok], "Expected test of pull request to be true"
+  ensure
+    ENV["CODECLIMATE_GITHUB_API_URL"] = old_base
```
pbrisbin: I just created a `stub_env` in `config_spec.rb`, might be worth moving to a shared helper to reuse here?

lib/cc/services/github_issues.rb
pbrisbin: Hmm, I guess I'm confused why you think this code wouldn't be able to call `CodeClimate.config`, it's all in the same Ruby process, right?

GordonDiggs:
@codeclimate/review Ready for another look

pbrisbin:
The change LGTM -- I think @brynary mentioned in chat that we need to make this field hidden before we can ship it, and I agree.

GordonDiggs:
@pbrisbin @brynary I thought the hidden-ness happened on the app side? In the way that we don't render a checkbox for the allow comments attribute I was going to do the same here

pbrisbin:
Sorry! Thought this was an App PR (that explains earlier confusion too). LGTM :)

brynary:
Let's make sure hidden is actually "protected" (as in, can't slip in bad values via unexpected form POST)



—Bryan Helmkamp, Founder, Code Climate

bryan@brynary.com / 646-379-1810 / @brynary




Typed with thumbs.

PR #82, opened 2015-09-15
don't remove details url for errored PRs

ABaldwinHunter:
@codeclimate/review 

Edit: now we can actually include the details link for errored PRs instead of deleting it when the snapshot is errored.


pbrisbin:
LGTM!

PR #84, opened 2015-11-02
Improve JIRA service

GordonDiggs:
- Add a description to the test issue. Organizations often have
  descriptions required for issues, meaning that our test fails to
  create an issue.
- Change URL of resulting hash. The URL returned from the API is the
  REST URL, and is not suitable for redirects.

@codeclimate/review


pbrisbin:
LGTM

PR #88, opened 2016-03-17
Specify what is considered true for config

wfleming:
A change in app has resulted in "0" getting written to the add_comment?
field in service records, which result in undesired comments on PRs.

I'm also seeing these value for the update_status? bit.

Digging deeper, I think we may have a general problem with how bools are being
stored here: I've tried configuring Virtus' bool casting, and looking deeper
into how we read & set things, and I'm not making much progress. This is
effectively a patch to fix affected customer behavior now, to give me some mroe
time to look more deeply into underyling problems.


wfleming:
cc @codeclimate/review

pbrisbin:
LGTM

PR #89, opened 2016-03-18
Services Type Coercion

wfleming:
Virtus depends on the type of attributes being the
appopriate `Axiom` class. This led to incorrect & unexpected results in
app, where other classes with the same name were getting resolved,
causing quiet failures.

I rolled back yesterday's temporary patch & added some small sanity tests as well.

cc @codeclimate/review @pbrisbin

### DIFF
--- a/lib/cc/service/config.rb
+++ b/lib/cc/service/config.rb
@@ -1,4 +1,4 @@
 class CC::Service::Config
-  include Virtus.model
+  include Virtus.model(coerce: true)
   include ActiveModel::Validations
 end
--- a/lib/cc/service/config.rb
+++ b/lib/cc/service/config.rb
@@ -1,4 +1,4 @@
 class CC::Service::Config
-  include Virtus.model(coerce: true)
+  include Virtus.model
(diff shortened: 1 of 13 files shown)


lib/cc/service/config.rb
```diff
@@ -1,4 +1,4 @@
 class CC::Service::Config
-  include Virtus.model
+  include Virtus.model(coerce: true)
```
pbrisbin: I'd probably ditch this since it's the default and wasn't the issue.

lib/cc/services/github_pull_requests.rb
```diff
@@ -123,6 +123,7 @@ def update_status(state, description)
   end
 
   def add_comment
+    $stderr.puts("WFDEBUG: config #{config.class} #{config.add_comment.inspect}\n\n")
```
GordonDiggs: Intentionally left?

wfleming:
@pbrisbin @GordonDiggs Updated.

pbrisbin:
LGTM

PR #91, opened 2016-04-15
Add bin tasks to streamline release process

ABaldwinHunter:
@codeclimate/review


GordonDiggs:
LGTM

wfleming:
The kind of wonky `prep-release` & `release` scripts are what they are on CLI because we have protected branches on that repo. We don't have protected branches on this repo, so this flow seems like overkill to me. I think it's actually making the release process more complicated. You should be able to just edit the `version.rb` file & run `rake release`?

GordonDiggs:
I actually vote we turn on protected branches here

bin/prep-release
```diff
@@ -0,0 +1,47 @@
...
+git add VERSION Gemfile.lock
+git commit -m "Release v$version"
+git push origin "$branch"
+
+branch_head=$(git rev-parse --short $branch)
+if command -v hub > /dev/null 2>&1; then
+  hub pull-request -F - <<EOF
+Release v$version
+
+https://github.com/codeclimate/codeclimate/compare/v$old_version...$branch_head
```
wfleming: Wrong repo URL here.

ABaldwinHunter:
I'm okay with turning on branch protection here.

bin/prep-release
ABaldwinHunter: ah good catch, thanks!

ABaldwinHunter:
My other thought is that I slightly like a `prep-release` script. Even though we typically merge anyway, I like semver feedback sometimes.

ABaldwinHunter:
Okay looks like there are some arguments for and against this change, so going to table for now.

dblandin:
@ABaldwinHunter Is this PR still active?

dblandin:
Going to close for now.

ABaldwinHunter:
👍 thanks

PR #92, opened 2016-05-24
Add GitHub issues service response message

dblandin:
This PR adds a message payload to override the default of "Success". The
override copy is actually used in the test but not implemented in the
`#create_issue` method.

### DIFF
--- a/lib/cc/services/github_issues.rb
+++ b/lib/cc/services/github_issues.rb
@@ -74,7 +74,8 @@ private
       {
         id: body["id"],
         number: body["number"],
-        url: body["html_url"]
+        url: body["html_url"],
+        message: "Issue <a href='#{body["html_url"]}'>##{body["number"]}</a> created.",
       }
     end
   end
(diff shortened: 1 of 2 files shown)


lib/cc/services/github_issues.rb
```diff
@@ -74,7 +74,8 @@ def create_issue(title, issue_body)
       {
         id: body["id"],
         number: body["number"],
-        url: body["html_url"]
+        url: body["html_url"],
+        message: "Issue <a href='#{body["html_url"]}'>##{body["number"]}</a> created.",
```
wfleming: Do any other services use HTML in their message fields?

dblandin: It looks like HTML is included within other message fields from the `#receive_test` payload. Here's [another example](https://github.com/codeclimate/codeclimate-services/blob/master/lib/cc/services/lighthouse.rb#L28). The content is likely what ends up on codeclimate.com when you click "Test integration". As to whether it was meant for other purposes, I'm not sure. A message of "Success" doesn't seem very useful.

Based on [bryan's comment](https://github.com/codeclimate/browser/pull/147#discussion_r64154644), it might make sense to widen this up. Or we could rely on client-side copy.

dblandin:
Going to close this for PR for now. We can revisit later if we want to make this change to all services.

PR #93, opened 2016-05-24
Turn GH PR update status to true by default

mrb:
Unclear why this is false, but I've gotten feedback a few times that it's confusing.


pbrisbin:
:+1:

PR #94, opened 2016-06-07
Provide default value for Asana notes

wfleming:
This value cannot be nil according to Asana's API.

cc @codeclimate/review

### DIFF
--- a/lib/cc/services/asana.rb
+++ b/lib/cc/services/asana.rb
@@ -54,7 +54,7 @@ class CC::Service::Asana < CC::Service
 
 private
 
-  def create_task(name, notes = nil)
+  def create_task(name, notes = "")
     params = generate_params(name, notes)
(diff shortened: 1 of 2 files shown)


GordonDiggs:
LGTM

lib/cc/services/asana.rb
```diff
@@ -54,7 +54,7 @@ def receive_vulnerability
 
 private
 
-  def create_task(name, notes = nil)
+  def create_task(name, notes = "(No details available)")
```
dblandin: Have you tried a blank string here instead? Does that also fail? Remembering the error, I thought it was more about the type of the attribute rather than content.

wfleming: You're right, an empty string is acceptable. Changed.

dblandin:
LGTM

PR #95, opened 2016-06-29
Improve refactor ticket title when there is no rating

wfleming:
The previous logic ended up saying "Refactor from a ?", which isn't great. This centralizes the title logic, which was implemented in a bunch of places, and emits a simpler title if there is no rating data.

cc @codeclimate/review


GordonDiggs:
👍

PR #97, opened 2016-07-18
Follow redirects for GH issue creation

wfleming:
I tracked this down related to a CX escalation. Issue creation URLs for
at least some repos are currently returning 307 statuses: [GH's docs][1]
make it fairly clear that we should not be changing the URL we start
with, but we should be following these. I was locally able to confirm
this fixed issue creation for an affected repo.

[1]: https://developer.github.com/v3/#http-redirects


pbrisbin:
Nice catch. LGTM modulo CC issues.

wfleming:
Our travis config seems to have gotten broken somehow. I've switched to Circle & turned off Travis rather than investigate that breakage too much, since we've been moving most of our repos to Circle.

PR #101, opened 2016-08-08
Show test coverage percent delta within commit status update

dblandin:
This commit changes the test coverage commit status description from:

`Test coverge for this commit: ##%`

to:

```
You're at X% test coverage
You're at X% test coverage (up X%)
You're at X% test coverage (down X%)
```

### DIFF
--- a/lib/cc/presenters/pull_requests_presenter.rb
+++ b/lib/cc/presenters/pull_requests_presenter.rb
@@ -12,6 +12,7 @@ module CC
         end
 
         @covered_percent = payload["covered_percent"]
+        @covered_percent_delta = payload["covered_percent_delta"]
       end
 
       def error_message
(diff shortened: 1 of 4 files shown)


pbrisbin:
I would run a screenshot by @mciarlo and/or @brynary 

I think space is tight, so WDYT of `+`/`-` instead of `increased`/`decreased`?

dblandin:
@mciarlo Any preference on copy?

2016-08-09-120741_721x275_scrot
2016-08-09-120751_717x273_scrot

mciarlo:
@dblandin A few suggestions. Feel free to pick your favorite or riff.

a) Your test coverage is at X% (up X%)
b) You're at X% test coverage (up X%)

dblandin:
@mciarlo Cool, thanks! I went with __b__, since _your test covearge_ reads a little odd to me. _Your tests_ or _your code_ sounds intuitive to me, but test coverage seems like a property of the tests or code, not something that's _yours_.

brynary:
I don't think we need a sentence structure I'd simply do:

**codeclimate/coverage** -- 87% test coverage (+1.2%)

Easy to read, and as a bonus it leaves more room for surprise and delight or other CTAs if we want to try one later.

dblandin:
> codeclimate/coverage -- 87% test coverage (+1.2%)

@brynary Works for me! I'll make that change.

brynary:
@dblandin thanks! Bonus question... Does it work if put emoji into this string? 🎉

dblandin:
> Bonus question... Does it work if put emoji into this string

Good question. Will check on that. If so, what kind of delta is emoji worthy? Is +0.01% worthy of a :tada:?

dblandin:
> Does it work if put emoji into this string

Confirmed that this does not work :disappointed:

pbrisbin:
LGTM

PR #102, opened 2016-08-08
Remove GH PR update_status and update_coverage_status checks

dblandin:
This PR removes configuration options to enable/disable commit status updates on active GitHub pull request integrations. These options made sense when we also supported commenting on pull requests, but since commit status updates are now the only mechanism we use, an active integration should also signify the intent to receive commit status updates.

To honor current configurations, we'll likely need a migration to set active integrations with `update_status=false` to an inactive state.


wfleming:
Code LGTM.

PR #104, opened 2016-08-11
Add GitLab MR support

GordonDiggs:
See individual commits for more info

@codeclimate/review


wfleming:
The duplication issue looks legit, except for the message in the rescue. Worth extracting?

wfleming:
One note, but LGTM!

GordonDiggs:
@wfleming I wanted to, but it's different enough that I don't think it would benefit from it

PR #105, opened 2016-08-12
Remove use of Numeric#positive? and Numeric#negative?

dblandin:
These two methods on `Numeric` were added in ruby `2.3` and we're
currently using this gem in services that run on ruby `2.2.0`.

Related ruby core feature request: https://bugs.ruby-lang.org/issues/11151


GordonDiggs:
LGTM

PR #109, opened 2016-08-22
Disallow setting the Gitlab context

GordonDiggs:
This does not need to be configurable, but does need to be on the config
object in order to match CC::PullRequests' interface

@codeclimate/review


wfleming:
LGTM. Why not allow it to be configurable, though, if we've found it useful to be configurable on other SCMs?

PR #110, opened 2016-08-22
Post welcome comments on pull requests

maxjacobson:
This PR introduces the ability for Code Climate to post a welcome comment on GitHub pull requests.

The body of the comment can be customized by the user, but the default is contained here.

The header (first line) of the comment *can't* be customized.

The footer of the comment *can't* be customized, and is meant to only be posted when the PR was opened by someone who administrates the repo, and therefore can also administrate the integration. It exists to remind them they can customize or disable the welcome comment.

A complete comment (including the admin-only footer) will look like this:

> Hey, **@maxjacobson**-- Since this is the first PR we've seen from you, here's some things you should know about contributing to codeclimate/codeclimate-services:
>
> * This repository is using Code Climate to automatically check for code quality issues.
> * You can see results for this analysis in the PR status below.
> * You can install [the Code Climate browser extension](https://codeclimate.com/browser) to see analysis without leaving GitHub.
>
> Thanks for your contribution!
>
> * * *
>
> Quick note: By default, Code Climate will post the above comment on the *first* PR it sees from each contributor. If you'd like to customize this message or disable this, go [here](https://codeclimate.com/repos/56e9708372d5ef36a3000f28/services/githubpullrequests/edit).

### DIFF
--- a/lib/cc/pull_requests.rb
+++ b/lib/cc/pull_requests.rb
@@ -1,8 +1,29 @@
 class CC::PullRequests < CC::Service
+  ABLE_TO_UPDATE_STATUS_MESSAGE = "Access token is valid".freeze
+  ABLE_TO_UPDATE_STATUS_AND_COMMENT_MESSAGE = "Access token is valid".freeze
+  ABLE_TO_UPDATE_STATUS_BUT_NOT_COMMENT_MESSAGE = "OAuth token requires 'repo' scope to post comments.".freeze
+
   def receive_test
     setup_http
 
-    receive_test_status
+    # this will raise an HTTPError or be successful:
+    able_to_update_status_response = check_if_able_to_update_status
+
+    if welcome_comment_implemented? && config.welcome_comment_enabled
+      able_to_comment_response = check_if_able_to_comment
+
+      ok, message =
+        if able_to_comment_response.fetch(:ok)
+          [true, ABLE_TO_UPDATE_STATUS_AND_COMMENT_MESSAGE]
+        else
+          [false, ABLE_TO_UPDATE_STATUS_BUT_NOT_COMMENT_MESSAGE]
+        end
+
+      able_to_comment_response.merge(able_to_update_status_response).
+        merge(ok: ok, message: message)
+    else
+      able_to_update_status_response
+    end
   end
 
   def receive_pull_request
@@ -69,7 +90,7 @@ class CC::PullRequests < CC::Service
     raise NotImplementedError
   end
 
-  def receive_test_status
+  def check_if_able_to_update_status
     url = base_status_url("0" * 40)
     params = { state: "success" }
     raw_post(url, params.to_json)
@@ -77,10 +98,10 @@ class CC::PullRequests < CC::Service
     if e.status == test_status_code
       {
         ok: true,
-        params: params.as_json,
-        status: e.status,
-        endpoint_url: url,
-        message: "Access token is valid",
+        able_to_update_status_params: params.as_json,
+        able_to_update_status_status: e.status,
+        able_to_update_status_endpoint_url: url,
+        message: ABLE_TO_UPDATE_STATUS_MESSAGE,
       }
     else
       raise
@@ -98,7 +119,8 @@ class CC::PullRequests < CC::Service
       state: state,
       target_url: @payload["details_url"],
     }
-    @response = service_post(status_url, params.to_json)
+    formatter = GenericResponseFormatter.new(http_prefix: :update_status_)
+    @response = service_post(status_url, params.to_json, formatter)
   end
 
   def status_url
@@ -111,6 +133,10 @@ class CC::PullRequests < CC::Service
 
   def setup_http
     raise NotImplementedError
+  end
+
+  def welcome_comment_implemented?
+    false
   end
 
   def commit_sha
--- a/lib/cc/pull_requests.rb
+++ b/lib/cc/pull_requests.rb
@@ -6,7 +6,6 @@ class CC::PullRequests < CC::Service
   def receive_test
     setup_http
 
-    # this will raise an HTTPError or be successful:
     able_to_update_status_response = check_if_able_to_update_status
 
     if welcome_comment_implemented? && config.welcome_comment_enabled
--- a/lib/cc/service/http.rb
+++ b/lib/cc/service/http.rb
@@ -1,4 +1,6 @@
 require "active_support/concern"
+require "cc/service/generic_response_formatter"
+require "cc/service/body_extracting_response_formatter"
 require "cc/service/response_check"
 
 module CC::Service::HTTP
@@ -21,20 +23,9 @@ module CC::Service::HTTP
     raw_get(url, body, headers, &block)
   end
 
-  def service_post(url, body = nil, headers = nil, &block)
-    block ||= ->(*_args) { Hash.new }
-    response = raw_post(url, body, headers)
-    formatted_post_response(response, url, body).merge(block.call(response))
-  end
-
-  def service_post_with_redirects(url, body = nil, headers = nil, &block)
-    block ||= ->(*_args) { Hash.new }
-    response = raw_post(url, body, headers)
-    if REDIRECT_CODES.include?(response.status)
-      response = raw_post(response.headers["location"], body, headers)
-    end
-
-    formatted_post_response(response, url, body).merge(block.call(response))
+  def service_post(url, body, formatter = CC::Service::GenericResponseFormatter.new)
+    response = raw_post(url, body)
+    formatter.post(url, body, response)
   end
 
   def raw_get(url = nil, params = nil, headers = nil)
@@ -46,9 +37,9 @@ module CC::Service::HTTP
     end
   end
 
-  def raw_post(url = nil, body = nil, headers = nil)
+  def raw_post(url = nil, body = nil)
     block = Proc.new if block_given?
-    http_method :post, url, body, headers, &block
+    http_method :post, url, body, nil, &block
   end
 
   def http_method(method, url = nil, body = nil, headers = nil)
@@ -88,15 +79,5 @@ module CC::Service::HTTP
   # Returns a String path.
   def ca_file
     @ca_file ||= ENV.fetch("CODECLIMATE_CA_FILE", File.expand_path("../../../../config/cacert.pem", __FILE__))
-  end
-
-  def formatted_post_response(response, url, body)
-    {
-      ok: response.success?,
-      params: body.as_json,
-      endpoint_url: url,
-      status: response.status,
-      message: "Success",
-    }
   end
 end
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -25,7 +25,6 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
 Thanks for your contribution!
       COMMENT
 
-
     validates :oauth_token, presence: true
   end
 
@@ -98,16 +97,16 @@ Thanks for your contribution!
     config.welcome_comment_enabled && @payload.fetch("first_contribution", false)
   end
 
-  HEADER_TEMPLATE = <<-HEADER
+  HEADER_TEMPLATE = <<-HEADER.freeze
 Hey, @%s-- Since this is the first PR we've seen from you, here's some things you should know about contributing to %s:
 
   HEADER
 
   def welcome_comment_markdown_header
-    HEADER_TEMPLATE % [@payload.fetch("author_username"), github_slug]
+    format HEADER_TEMPLATE, @payload.fetch("author_username"), github_slug
   end
 
-  ADMIN_ONLY_FOOTER_TEMPLATE = <<-FOOTER
+  ADMIN_ONLY_FOOTER_TEMPLATE = <<-FOOTER.freeze
 
 * * *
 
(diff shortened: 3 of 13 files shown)


lib/cc/services/github_pull_requests.rb
```diff
@@ -73,4 +89,48 @@ def response_includes_repo_scope?(response)
...
+    end
+  end
+
+  def post_welcome_comment
+    pull_request_number = @payload.fetch("number")
+    comments_url = "#{config.base_url}/repos/#{github_slug}/issues/#{pull_request_number}/comments"
+
+    params = { body: welcome_comment_markdown }
+
+    # FIXME: is this appropriate? when we receive the pending event, we used to just update the status for the commit,
```
wfleming: Personal opinion: if we have to pick one, the commit status is more important.

But it sounds like maybe we need to support multiple requests/responses? Do you know where/how our `ServiceEvent` records are created in response to these actions?

Rebuilding a lot of infrastructure to support multiple responses may not be reasonable right now. I'm not sure.

pull_request_test.rb
```diff
@@ -28,16 +28,19 @@ def call
   end
 end
 
+token = ENV["OAUTH_TOKEN"] || File.read("OAUTH_TOKEN").chomp
```
wfleming: Why put it in a file?

lib/cc/services/github_pull_requests.rb
maxjacobson: hey, sorry for the slow response here

I believe this is the place: https://github.com/codeclimate/app/blob/ee8f89556a2af398afdfe2b4bb771c1a0ba09d1f/app/models/service.rb#L193-L200

(there might be other places, too)

I think going forward with returning the status update's response is OK for me

wfleming: I kinda wish we could keep both. Without the ServiceEvent, we won't have much debugging info to look at in prod for how this is actually performing.

maxjacobson: OK my new plan is to merge the two network responses into one service event that reflects the success of both. It will be `ok: true` when both requests succeeded, and it will have the [metadata](https://github.com/codeclimate/codeclimate-services/blob/e0e48d49bb3f268ee07d094b1101799ad2247bee/lib/cc/service/http.rb#L93-L100) from both responses for debugging purposes

wfleming: That sounds reasonable to me.

In the event of one failure/one pass will the merged metadata be sufficient to know which is which?

maxjacobson: Digging into / implementing this more reveals this current understanding of the situation:

we can merge the responses when both succeed, but when either fails, the handling is bubbled up to [totally different code](https://github.com/codeclimate/codeclimate-services/blob/e0e48d49bb3f268ee07d094b1101799ad2247bee/lib/cc/service/invocation/with_error_handling.rb#L11-L20), which will let us know which one failed, and won't be aware that one may have succeeded. That might be OK, because updating a status a second time is harmless (right?) and we have a check to make sure we don't double-comment.

pull_request_test.rb
maxjacobson: The idea here was that it's a little easier IMO to manage this file in a local file rather than an environment variable in my global dotfiles or in 1password, kind of like our `.env.development` files in other repos. I couldn't actually figure out how to use this file to test anything, so I'm likely to just undo these changes, unless I can figure out how to use this.

maxjacobson:
@codeclimate/review Hey folks! I'm ready for feedback on this. I think this change is mostly finished. Note that some of this came by referencing [this commit](https://github.com/codeclimate/codeclimate-services/commit/5a2dd0b591d2bb33edf4946ad90e36cbcf927137)

Particularly, if someone would like to volunteer to help me get my local setup working to the point where I can test this change integrated with the system, I would be very grateful.

NOTES.txt
```diff
@@ -0,0 +1,10 @@
+Need app to start emitting in pull request payloads:
```
wfleming: Was this file intended to be left?

lib/cc/pull_requests.rb
```diff
@@ -2,7 +2,28 @@ class CC::PullRequests < CC::Service
   def receive_test
     setup_http
 
-    receive_test_status
+    response = receive_test_status
```
wfleming: Don't we want to do this for the PR receive, not the test coverage receive event?

lib/cc/pull_requests.rb
```diff
@@ -113,6 +134,10 @@ def setup_http
     raise NotImplementedError
   end
 
+  def welcome_comment_implemented?
+    raise NotImplementedError
```
wfleming: I think it would be fine to have this be `false` in the root class rather than forcing every class to specify.

lib/cc/pull_requests.rb
```diff
@@ -2,7 +2,28 @@ class CC::PullRequests < CC::Service
...
 
-    receive_test_status
+    response = receive_test_status
+
+    if welcome_comment_implemented? && config.welcome_comment_enabled
+      welcome_comment_response = receive_test_welcome_comment
+      if response[:ok] && welcome_comment_response[:ok]
+        {
+          ok: true,
+          message: "Access token is valid",
```
wfleming: There are reasons requests can fail other than access tokens. Maybe something like `"Pull Request Status Set & Welcome Comment Posted"`?

lib/cc/pull_requests.rb
```diff
@@ -2,7 +2,28 @@ class CC::PullRequests < CC::Service
...
+          requests: [
+            response.slice(:params, :status, :endpoint_url),
```
wfleming: Any reason not to include all the params? What are the other ones?

lib/cc/pull_requests.rb
```diff
@@ -2,7 +2,28 @@ class CC::PullRequests < CC::Service
...
+            welcome_comment_response.slice(:params, :status, :endpoint_url),
+          ]
+        }
+      else
+        {
+          ok: false,
+          message: welcome_comment_response[:message],
```
wfleming: We should include the full response info of both here as well.

lib/cc/services/github_pull_requests.rb
```diff
@@ -73,4 +89,82 @@ def response_includes_repo_scope?(response)
...
+
+  def welcome_comment_implemented?
+    true
+  end
+
+  def should_post_welcome_comment?
+    config.welcome_comment_enabled && @payload.fetch("first_contribution", false)
+  end
+
+  ALREADY_COMMENTED_PATTERN = /Since this is the first PR/
```
wfleming: Hmm. Don't we externally decide if we've already sent this and set it when sending the service event? Are you worried about race conditions?

maxjacobson: I was more concerned about our retrying logic. I was thinking specifically if our status update fails, but the comment succeeds, we might retry and double-comment. With a little more context of the code, I actually don't think that will happen, because we'll fail hard on the status update failure. Hypothetically, if we were to add a third component of this service in the future, which takes place after the comment, and *that* were to fail, we might retry and double-comment (without this check).

wfleming: I think this then falls under YAGNI. That hypothetical third component doesn't exist, and if it ever does I think it'll mean we need to consider some deeper refactoring here.

lib/cc/pull_requests.rb
maxjacobson: If I'm understanding this correctly, this `receive_test` method is called when the user uses the "test integration" button from CC, and it's meant to verify that the provided oauth token is going to be sufficient for the service to do everything it's supposed to be able to do. So previously, it checked that it was able to update statuses, and now it might additionally check if it will be able to post comments.

lib/cc/pull_requests.rb
maxjacobson: At this point, we haven't posted the comment or updated the status yet, but verified that with the provided token, we'll be able to (this is for the "Test Service" button). Given, that I think the wording makes sense but am able to be convinced otherwise.

I think this code is kind of confusing currently... going to spend some time trying to make it more straight-forward now

maxjacobson:
@wfleming I took your above suggestions and simplified some other things... will appreciate another round of 👀 (prob next week 😄)

lib/cc/pull_requests.rb
wfleming: Ah, sorry, I confused myself, and was thinking of the new PR integration support for test coverage. You're quite right, ignore me.

lib/cc/pull_requests.rb
wfleming: Same as I said above: I was confused, sorry, you're totally right.

lib/cc/pull_requests.rb
```diff
@@ -1,8 +1,28 @@
 class CC::PullRequests < CC::Service
+  ABLE_TO_UPDATE_STATUS_MESSAGE = "Access token is valid"
+  ABLE_TO_UPDATE_STATUS_AND_COMMENT_MESSAGE = "Access token is valid"
+  ABLE_TO_UPDATE_STATUS_BUT_NOT_COMMENT_MESSAGE = "OAuth token requires 'repo' scope to post comments."
```
wfleming: :lipstick: missing space between the constants and the method.

lib/cc/pull_requests.rb
```diff
@@ -1,8 +1,28 @@
 class CC::PullRequests < CC::Service
+  ABLE_TO_UPDATE_STATUS_MESSAGE = "Access token is valid"
```
wfleming: :lipstick: this is probably a missing/different rubocop config in this project, but we prefer to `freeze` mutable objects assigned to constants.

lib/cc/pull_requests.rb
```diff
@@ -1,8 +1,28 @@
...
+  ABLE_TO_UPDATE_STATUS_AND_COMMENT_MESSAGE = "Access token is valid"
+  ABLE_TO_UPDATE_STATUS_BUT_NOT_COMMENT_MESSAGE = "OAuth token requires 'repo' scope to post comments."
   def receive_test
     setup_http
 
-    receive_test_status
+    # this will raise an HTTPError or be successful:
```
wfleming: Is the comment going to be helpful in the long term?

lib/cc/pull_requests.rb
```diff
@@ -1,8 +1,28 @@
...
+    able_to_update_status_response = check_if_able_to_update_status
+
+    if welcome_comment_implemented? && config.welcome_comment_enabled
```
wfleming: The structure here feels confusing in that we're treating the status test & the comment test quite differently. Can we make these more similar? Is it problematic that we're not raising an exception for comment failure the same as status failure? This might be a good problem to pair on.

lib/cc/service/http.rb
```diff
@@ -21,10 +21,10 @@ def service_get(url = nil, body = nil, headers = nil, &block)
     raw_get(url, body, headers, &block)
   end
 
-  def service_post(url, body = nil, headers = nil, &block)
+  def service_post(url, body = nil, headers = nil, response_prefix = nil, &block)
```
wfleming: I don't think this prefixing stuff belongs here in the HTTP layer: it's really a concern you have with presentation of results, not HTTP. It seems like there's another concept hiding here.

lib/cc/pull_requests.rb
maxjacobson: Yep, we just infer it currently, but should use the house style. Will update. In the meantime: https://github.com/codeclimate/codeclimate-services/pull/111

lib/cc/service/http.rb
maxjacobson: I think it's a big-ish refactor to remove the response-formatting from the http code. I'm open to doing it, though, as it will make it easier to make changes like this one which accommodate multiple request metadata in one service response

lib/cc/pull_requests.rb
maxjacobson: Yes, I'd like to take you up on that 😄

lib/cc/service/http.rb
maxjacobson: I went forward with extracting some response formatter classes... the nice thing is that we can just inject the formatter class, and so the http layer isn't concerned with formatting the response, it just uses the injected formatter to do that, which gives us flexiblity to have a few different kinds of formatters, which enables prefixing sometimes... WDYT?

lib/cc/pull_requests.rb
maxjacobson: If we want to make them more similar, I think we could make the status check more like the comment check. The comment check makes a request to confirm that the provided auth token has been granted the appropriate permissions which it will need to post a comment. The status check, on the other hand, attempts to update a status for a nonexistent commit (one with a SHA of `0000000000000000000000000000000000000000`) and checks the status code of the response to make sure it indicates "you're allowed to, but that commit doesn't exist" rather than "you're not allowed to". The downside to this approach is that, I believe, it will retry 3 times (which is a bit slow/futile) and doesn't give as actionable advice (telling them which permission was missing from the oauth token). Changing this behavior would mean changing both the github and the gitlab integration. At this point in time I'm inclined to tolerate the discrepancy but will get another opinion on it

lib/cc/service/http.rb
```diff
@@ -21,20 +23,9 @@ def service_get(url = nil, body = nil, headers = nil, &block)
     raw_get(url, body, headers, &block)
   end
 
-  def service_post(url, body = nil, headers = nil, &block)
-    block ||= ->(*_args) { Hash.new }
-    response = raw_post(url, body, headers)
-    formatted_post_response(response, url, body).merge(block.call(response))
-  end
-
-  def service_post_with_redirects(url, body = nil, headers = nil, &block)
```
wfleming: Why was this removed? Is there a replacement for it?

maxjacobson:
(Going to close and re-open a simpler version)

PR #111, opened 2016-08-29
Add RuboCop config from our styleguide

maxjacobson:
We were using the less opinionated version that CC infers based on the project.

(Inspired by this comment: https://github.com/codeclimate/codeclimate-services/pull/110#discussion_r76623698)

### DIFF
--- a/bin/bundler
+++ b/bin/bundler
@@ -6,11 +6,11 @@
 # this file is here to facilitate running it.
 #
 
-require 'pathname'
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path("../../Gemfile",
+require "pathname"
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",
   Pathname.new(__FILE__).realpath)
 
-require 'rubygems'
-require 'bundler/setup'
+require "rubygems"
+require "bundler/setup"
 
-load Gem.bin_path('codeclimate-services', 'bundler')
+load Gem.bin_path("codeclimate-services", "bundler")
--- a/lib/cc/service/helper.rb
+++ b/lib/cc/service/helper.rb
@@ -1,6 +1,6 @@
 module CC::Service::Helper
-  GREEN_HEX = "#38ae6f"
-  RED_HEX   = "#ed2f00"
+  GREEN_HEX = "#38ae6f".freeze
+  RED_HEX = "#ed2f00".freeze
 
   def repo_name
     payload["repo_name"]
@@ -50,5 +50,4 @@ module CC::Service::Helper
     raise NotImplementedError,
       "Event-specific helpers must define #{__method__}"
   end
-
 end
--- a/lib/cc/service/response_check.rb
+++ b/lib/cc/service/response_check.rb
@@ -5,9 +5,9 @@ class CC::Service
 
     def initialize(message, env)
       @response_body = env[:body]
-      @status        = env[:status]
-      @params        = env[:params]
-      @endpoint_url  = env[:url].to_s
+      @status = env[:status]
+      @params = env[:params]
+      @endpoint_url = env[:url].to_s
 
       super(message)
     end
@@ -25,7 +25,7 @@ class CC::Service
       end
     end
 
-  private
+    private
 
     def error_message(env)
       # We only handle Jira (or responses which look like Jira's). We will add
@@ -37,6 +37,5 @@ class CC::Service
       end
     rescue JSON::ParserError
     end
-
   end
 end
--- a/lib/cc/services/asana.rb
+++ b/lib/cc/services/asana.rb
@@ -14,7 +14,7 @@ class CC::Service::Asana < CC::Service
     validates :workspace_id, presence: true
   end
 
-  ENDPOINT = "https://app.asana.com/api/1.0/tasks"
+  ENDPOINT = "https://app.asana.com/api/1.0/tasks".freeze
 
   self.title = "Asana"
   self.description = "Create tasks in Asana"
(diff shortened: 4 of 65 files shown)


GordonDiggs:
A lot of the new issues can be autofixed with rubocop directly

maxjacobson:
@GordonDiggs True! I was thinking it might make sense to follow the process where we don't try and fix things in one big clump, but rather over time as the files are edited anyway. I did fix one category of issue because it was the one pointed out in that comment, but wasn't planning to fix everything. Do you have an opinion about how much would be appropriate to change at this point in time?

GordonDiggs:
I think when we update the styleguide on a repo it makes sense to fix as much of it as makes sense. IMO all the stuff that can be autofixed is a no brainer. The only things I wouldn't fix would be the length-related issues

maxjacobson:
@GordonDiggs -- cool, updated. Lots of changes now. Tests are still green. I want to review a little more closely to make sure the autocorrect didn't mangle anything... feel free to do the same!

maxjacobson:
OK I combed thru the autocorrect and I don't think it broke anything, and snapped a bunch of things into a consistent place. Ready for another look @codeclimate/review

bin/bundler
```diff
@@ -6,11 +6,11 @@
 # this file is here to facilitate running it.
 #
 
-require 'pathname'
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path("../../Gemfile",
-  Pathname.new(__FILE__).realpath)
+require "pathname"
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",
+                                           Pathname.new(__FILE__).realpath)
```
pbrisbin: This isn't actually how we do this.

```rb
ENV["BUNDLE_GEMFILE"] ||= File.expand_path(
  "../../Gemfile",
  Pathname.new(__FILE__).realpath
)
```

Is something off in our rubcop config, that it's allowed this?

lib/cc/service/helper.rb
```diff
@@ -48,7 +48,6 @@ def changed
 
   def improved?
     raise NotImplementedError,
-      "Event-specific helpers must define #{__method__}"
+          "Event-specific helpers must define #{__method__}"
```
pbrisbin: This seems wrong

lib/cc/service/response_check.rb
```diff
@@ -19,13 +19,13 @@ class ResponseCheck < Faraday::Response::Middleware
     def on_complete(env)
       if ErrorStatuses === env[:status]
         message = error_message(env) ||
-          "API request unsuccessful (#{env[:status]})"
+                  "API request unsuccessful (#{env[:status]})"
```
pbrisbin: This too. I can understand why it'd get flagged, but the fix of pushing it over further doesn't seem right.

lib/cc/services/asana.rb
```diff
@@ -5,16 +5,16 @@ class Config < CC::Service::Config
     attribute :workspace_id, Axiom::Types::String, label: "Workspace ID"
 
     attribute :project_id, Axiom::Types::String, label: "Project ID",
-      description: "(optional)"
+                                                 description: "(optional)"
 
     attribute :assignee, Axiom::Types::String, label: "Assignee",
-      description: "Assignee email address (optional)"
+                                               description: "Assignee email address (optional)"
```
pbrisbin: I think our preferred way to multi-line-format these is

```rb
attribute \
  :assignee
  ...String,
  label: "x",
  description: "y"
```

test/formatters/snapshot_formatter_test.rb
```diff
@@ -6,40 +6,37 @@ def described_class
...
-    f = CC::Formatters::SnapshotFormatter::Sample.new({
-       "new_constants" => [{"name" => "foo", "to" => {"rating" => "A"}}, {"name" => "bar", "to" => {"rating" => "A"}}],
-       "changed_constants" => [
-         {"from" => {"rating" => "B"}, "to" => {"rating" => "C"}},
-         {"from" => {"rating" => "D"}, "to" => {"rating" => "D"}},
-         {"from" => {"rating" => "D"}, "to" => {"rating" => "D"}},
-         {"from" => {"rating" => "A"}, "to" => {"rating" => "B"}},
-         {"from" => {"rating" => "C"}, "to" => {"rating" => "B"}}
-       ]})
+    f = CC::Formatters::SnapshotFormatter::Sample.new("new_constants" => [{ "name" => "foo", "to" => { "rating" => "A" } }, { "name" => "bar", "to" => { "rating" => "A" } }],
```
pbrisbin:
```rb
f = ...Sample.new(
  "new_constants" => [...],
  "changed_constants" => [
    { ...
    { ...
  ]
)
```

pbrisbin:
You could've totally considered this a "merge on green", but since you asked I went ahead and picked the nits I saw. Note that there are a lot of instances of same issue (particularly multi-line method calls), and I tried to only comment on the first I saw.

maxjacobson:
@pbrisbin thanks! I'll see if I can tighten those up. Unfortunately I'm not an admin on this repo yet, so I can't merge around the status check, so I may be pinging you soon to help merge this (or feel free to add me 😄)

pbrisbin:
Cool. You can also use shipbot.

```
merge [PR URL]
  "Mercy merge" a PR using protected branches
  Ensure you've cleared this publically with at least one reviewer, or it's
  "extenuating circumstances" (e.g. an incident)
```

maxjacobson:
@pbrisbin ah, sweet. OK. I updated the config to be more in line with our current style. There's one bug in rubocop that prevents it from addressing all of your comments (I filed an issue), but it's a lot closer. If it looks good to you, I'll bring those changes (the stuff in `.rubocop.yml`) over to styleguide, and then re-sync the `base_rubocop.yml` with styleguide

pbrisbin:
Sounds good to me

PR #114, opened 2016-09-02
Post welcome comments on pull requests

maxjacobson:
This PR introduces the ability for Code Climate to post a welcome comment on GitHub pull requests.

The body of the comment can be customized by the user, but the default is contained here.

The header (first line) of the comment *can't* be customized.

The footer of the comment *can't* be customized, and is meant to only be posted when the PR was opened by someone who administrates the repo, and therefore can also administrate the integration. It exists to remind them they can customize or disable the welcome comment.

A complete comment (including the admin-only footer) will look like this:

> Hey, **@maxjacobson** -- Since this is the first PR we've seen from you, here's some things you should know about contributing to codeclimate/codeclimate-services:
>
> * This repository is using Code Climate to automatically check for code quality issues.
> * You can see results for this analysis in the PR status below.
> * You can install [the Code Climate browser extension](https://codeclimate.com/browser) to see analysis without leaving GitHub.
>
> Thanks for your contribution!
>
> * * *
> Quick note: By default, Code Climate will post the above comment on the *first* PR it sees from each contributor. If you'd like to customize this message or disable this, go [here](https://codeclimate.com/repos/56e9708372d5ef36a3000f28/services/githubpullrequests/edit).

### DIFF
--- a/lib/cc/pull_requests.rb
+++ b/lib/cc/pull_requests.rb
@@ -1,8 +1,15 @@
 class CC::PullRequests < CC::Service
+  VALID_TOKEN_MESSAGE = "Access token is valid.".freeze
+  CANT_UPDATE_STATUS_MESSAGE = "Access token is invalid - can't update status.".freeze
+
   def receive_test
     setup_http
 
-    receive_test_status
+    if able_to_update_status?
+      { ok: true, message: VALID_TOKEN_MESSAGE }
+    else
+      { ok: false, message: CANT_UPDATE_STATUS_MESSAGE }
+    end
   end
 
   def receive_pull_request
@@ -73,19 +80,13 @@ class CC::PullRequests < CC::Service
     raise NotImplementedError
   end
 
-  def receive_test_status
-    url = base_status_url("0" * 40)
-    params = { state: "success" }
-    raw_post(url, params.to_json)
+  def able_to_update_status?
+    raw_post(base_status_url("0" * 40), { state: "success" }.to_json)
   rescue CC::Service::HTTPError => e
     if e.status == test_status_code
-      {
-        ok: true,
-        params: params.as_json,
-        status: e.status,
-        endpoint_url: url,
-        message: "Access token is valid",
-      }
+      true
+    elsif (400..499).cover?(e.status)
+      false
     else
       raise
     end
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -59,7 +59,7 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
     }
   end
 
-  def receive_pull_request_welcome_comment
+  def receive_pull_request_opened
     return unless config.welcome_comment_enabled
 
     setup_http
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -61,6 +61,7 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
 
   def receive_pull_request_opened
     return unless config.welcome_comment_enabled
+    return unless payload.fetch("authors_first_contribution")
 
     setup_http
 
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -169,8 +169,12 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
     GitHubPullRequestsWelcomeCommentPresenter.new(@payload, config).welcome_message
   end
 
+  def welcome_comment_enabled?
+    [true, "1"].include?(config.welcome_comment_enabled)
+  end
+
   def able_to_post_comments?
-    if config.welcome_comment_enabled
+    if welcome_comment_enabled?
       able_to_comment?
     end
   end
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -1,4 +1,5 @@
 require "cc/presenters/pull_requests_presenter"
+require "cc/presenters/github_pull_requests_welcome_comment_presenter"
 
 class CC::Service::GitHubPullRequests < CC::PullRequests
   class Config < CC::Service::Config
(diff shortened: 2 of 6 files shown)


lib/cc/services/github_pull_requests.rb
```diff
@@ -13,13 +13,35 @@ class Config < CC::Service::Config
...
+Thanks for your contribution!
+    COMMENT
 
     validates :oauth_token, presence: true
   end
 
   self.title = "GitHub Pull Requests"
   self.description = "Update pull requests on GitHub"
 
+  def receive_pull_request_welcome_comment
```
wfleming: Since this event is only for PR integrations, do we need `pull_request` as part of the event name? Seems redundant.

lib/cc/services/github_pull_requests.rb
```diff
@@ -13,13 +13,35 @@ class Config < CC::Service::Config
       label: "Github Context",
       description: "The integration name next to the pull request status",
       default: "codeclimate"
+    attribute :welcome_comment_enabled, Axiom::Types::Boolean,
+      label: "Welcome comment enabled?",
+      description: "Should Code Climate post a welcome comment on pull requests?",
```
wfleming: Might want to get some copy team opinions here, but personally based on what I see in other integrations, I think we could go with something briefer. Maybe `"Post a welcome comment?"`.

lib/cc/pull_requests.rb
```diff
@@ -1,8 +1,34 @@
...
+  CANT_UPDATE_STATUS_MESSAGE = "Access token is invalid - can't update status.".freeze
+  INVALID_TOKEN_MESSAGE = "Access token is invalid.".freeze
+
   def receive_test
     setup_http
 
-    receive_test_status
+    able_to_update_status = able_to_update_status?
+
+    able_to_post_comments =
```
wfleming: I continue to not entirely trust extension test coverage behavior. I saw a test that hits this method, didn't I? But the extension shows most lines in here as uncovered.

lib/cc/services/github_pull_requests.rb
```diff
@@ -66,11 +88,64 @@ def github_slug
...
+  ADMIN_ONLY_FOOTER_TEMPLATE = <<-FOOTER.freeze
+* * *
+Quick note: By default, Code Climate will post the above comment on the *first* PR it sees from each contributor. If you'd like to customize this message or disable this, go [here](%s).
+  FOOTER
+
+  def welcome_comment_footer
+    format ADMIN_ONLY_FOOTER_TEMPLATE, @payload.fetch("pull_request_integration_edit_url")
+  end
+
+  def author_is_site_admin?
```
wfleming: naming nit, but this is really checking for if they can admin the repo on cc.com, right? To me "site admin" makes me think of super admins, e.g. our team. Would `author_can_admin_repo?` be accurate here?

wfleming:
I have a couple nits here, and the CC issues seem worth addressing, but I really like the direction overall!

lib/cc/pull_requests.rb
maxjacobson: Hm... yea, that's not great. I ran simplecov locally and I see the same results, so I believe the extension is faithfully rendering them at least. Worth digging into, for sure

pbrisbin: > I ran simplecov locally and I see the same results,

That's good news. Worth investigating, but I think this definitively rules out bugs on our side.

dblandin:
@maxjacobson Did you consider adding this functionality directly on the `GitHubPullRequests` integration? Seems like we're guarding too much on the `welcome_comment_implemented?` check in the parent class when that could be pushed down. I don't see much value in the reuse right now.

We also have that kind of check built into `CC::Service,` so if a concrete service doesn't implement `receive_pull_request_welcome_comment`, it simply shouldn't receive the event.

lib/cc/services/github_pull_requests.rb
maxjacobson: :thumbsup: I'll make that change now and plan to share screenshots with copy folks when they're ready to get a coat of paint

lib/cc/services/github_pull_requests.rb
maxjacobson: :thumbsup: I'll make that change. Was planning to pull this information from the github payload (which has a boolean key called `site_admin`) as a proxy for whether they'll be able to administrate the cc repo, but we actually will have an opportunity to check in app if they can administrate the cc repo before sending this event

maxjacobson:
@dblandin good point, I like that idea. Updating

lib/cc/pull_requests.rb
wfleming: This repo is still using ancient `Test::Unit`. I suspect that's the reason. I'm looking into migrating to rspec.

lib/cc/services/github_pull_requests.rb
```diff
@@ -13,13 +14,53 @@ class Config < CC::Service::Config
...
+  }.freeze
+
+  # Override:
+  def receive_test
+    setup_http
+
+    tests = [able_to_update_status?, able_to_post_comments?]
+
+    {
+      ok: tests.compact.all?,
```
wfleming: neat.

wfleming:
This LGTM. It's difficult to be confident about it without having built the other pieces that talk to it though (e.g. from worker). Do you want to release this or hold on to it until we can test in conjunction with other changes?

lib/cc/pull_requests.rb
```diff
@@ -1,8 +1,15 @@
...
+
   def receive_test
     setup_http
 
-    receive_test_status
+    if able_to_update_status?
+      { ok: true, message: VALID_TOKEN_MESSAGE }
+    else
+      { ok: false, message: CANT_UPDATE_STATUS_MESSAGE }
+    end
```
dblandin: This seems less useful than the default behavior which includes the external error message: https://github.com/codeclimate/codeclimate-services/blob/master/lib/cc/service/invocation/with_error_handling.rb

What do you think about reducing this diff a bit further or is there a functional reason for these changes?

lib/cc/services/github_pull_requests.rb
```diff
@@ -13,13 +14,53 @@ class Config < CC::Service::Config
...
+    [true, nil] => VALID_TOKEN_MESSAGE,
+    [false, nil] => CANT_UPDATE_STATUS_MESSAGE,
+    [true, false] => CANT_POST_COMMENTS_MESSAGE,
+    [false, true] => CANT_UPDATE_STATUS_MESSAGE,
+    [false, false] => INVALID_TOKEN_MESSAGE,
+  }.freeze
+
+  # Override:
+  def receive_test
+    setup_http
```
dblandin: Would something like this work, leveraging the default behavior which tests token validity?

```ruby
def receive_test
  super

  if able_to_post_comments?
    { ok: true, message: "Access token is valid." }
  else
    { ok: false, message: "Access token does not have full repo scope." }
  end
end
```

test/github_pull_requests_test.rb
```diff
@@ -90,6 +90,24 @@ def test_pull_request_status_test_success
     assert receive_test({}, github_slug: "pbrisbin/foo")[:ok], "Expected test of pull request to be true"
   end
 
+  def test_pull_request_status_test_success_and_comment_success
+    @stubs.post("/repos/pbrisbin/foo/statuses/#{"0" * 40}") { |_env| [422, {}, ""] }
```
wfleming: You're gonna hit conflicts rebasing of course. Some tips for the conversion:

* `@stubs` becomes `http_stubs`
* `receive` becomes `service_receive`
* All the `assert` forms should be transformed to their equivalent `expect` form.

lib/cc/pull_requests.rb
wfleming: The test status is dependent on two different HTTP requests. Are you suggesting we should only show the first one that failed? Or combine them if they both failed?

I think the higher level messaging is actually useful: we're telling the user directly that this token needs to be able to set commit statuses, and also post comments. Error messages straight from the source are also helpful, of course. Maybe we should combine these and make the message strings format strings we can insert the HTTP response message into.

dblandin: I agree with you on this for the `GitHubPullRequests` functionality. See this [comment](https://github.com/codeclimate/codeclimate-services/pull/114#discussion_r77405047).

But the default behavior here only requires the token to be valid for setting commit statuses, not for commenting, so only one HTTP request. I think that the default behavior which uses the response for the error message is more helpful in this shared class.

lib/cc/services/github_pull_requests.rb
maxjacobson: The current implementation of `PullRequest#receive_test` (on master) will raise an unhandled `HTTPError` when we call `super`, so we wouldn't proceed and then *also* test the ability-to-comment when the status check test fails... are you saying that might be fine?

test/github_pull_requests_test.rb
maxjacobson: thanks, super helpful! there were enough changes to existing tests that I just sort of ported this file over manually from the version on this branch, which led to some scenario description strings being a bit different than you did, but I think it's pretty much otherwise the same

maxjacobson:
> Do you want to release this or hold on to it until we can test in conjunction with other changes?

Am inclined to hold until the other pieces are ready and until after talking thru the design decisions a bit more with Devon 😄

maxjacobson:
Putting this feature on ice 🍧

PR #115, opened 2016-09-02
Respect redirects when creating GH issues

wfleming:
This follow-redirect behavior was implemented in #97 specifically to
account for some GH API behavior, but as Max very kindly pointed out
recently, we weren't actually using it where we had intended to.

Now we are using it.

cc @codeclimate/review

### DIFF
--- a/lib/cc/services/github_issues.rb
+++ b/lib/cc/services/github_issues.rb
@@ -67,7 +67,7 @@ class CC::Service::GitHubIssues < CC::Service
     http.headers["User-Agent"] = "Code Climate"
 
     url = "#{config.base_url}/repos/#{config.project}/issues"
-    service_post(url, params.to_json) do |response|
+    service_post_with_redirects(url, params.to_json) do |response|
       body = JSON.parse(response.body)
       {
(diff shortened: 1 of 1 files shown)


pbrisbin:
LGTM

The red treatment reminded me: do we have some issue in this repo where test coverage isn't being handled correctly? I'm pretty sure that line, and others I've seen, is covered.

wfleming:
> The red treatment reminded me: do we have some issue in this repo where test coverage isn't being handled correctly? I'm pretty sure that line, and others I've seen, is covered.

I just mentioned the same thing on #110. Maybe it is specific to how test coverage is being reported on this repo: [on the site](https://codeclimate.com/github/codeclimate/codeclimate-services/lib/cc/services/github_issues.rb) it also looks mostly uncovered, though we do have tests for these methods.

Maybe this is an incompatibility between our test reporter and the version of minitest being used or something. I'll take a look.

maxjacobson:
(Actually https://github.com/codeclimate/codeclimate-services/pull/114/files#r77356061 not #110) (sorry) (:smile:)

PR #116, opened 2016-09-02
Transform test suite to RSpec

wfleming:
We noticed we weren't getting correct test coverage results on this repo. Digging into it, this test  suite was still on `Test::Unit`, and simplecov appears to not work properly with that anymore.

I've converted the suite to RSpec, which brings it more in line with our other projects. Locally running SimpleCov also shows that it appears to fix the inaccurate coverage results as well.

This conversion was done mostly with a bunch of local `sed` scripting & manually fixing cases I couldn't handle automatically, so there's probably still some messy bits. I personally don't think we need to get this perfect in every little bit of the test suite. Tests pass with no warnings, so I think it's pretty good.


wfleming:
cc @codeclimate/review

maxjacobson:
> We noticed we weren't getting correct test coverage results on this repo. Digging into it, this test suite was still on Test::Unit, and simplecov appears to not work properly with that anymore.

wonder if it's worth digging into that and possibly printing a warning from our test reporter and/r refusing to send payloads if we detect they're using `Test::Unit`

(looking at the content of the conversion now)

Rakefile
```diff
@@ -1,11 +1,6 @@
...
-require "rake/testtask"
+require "rspec/core/rake_task"
 
-Rake::TestTask.new do |t|
-  t.libs.push "lib"
-  t.libs.push "test"
-  t.test_files = FileList["test/**/*_test.rb"]
-  t.verbose = true
-end
+RSpec::Core::RakeTask.new(:spec)
```
maxjacobson: if you want to maintain the behavior where we print ruby compiler warnings, this should be changed to:

```ruby
RSpec::Core::RakeTask.new(:spec) do |task|
  task.ruby_opts = "-W"
end
```

wfleming:
> wonder if it's worth digging into that and possibly printing a warning from our test reporter and/r refusing to send payloads if we detect they're using Test::Unit

Not sure it's worth it. `Test::Unit` hasn't been the default way to write Ruby tests for several years.

Rakefile
wfleming: Good idea, done.

maxjacobson:
This LGTM -- now my other PR will have more accurate coverage

Nice:

screenshot 2016-09-02 16 10 09

PR #117, opened 2016-09-14
WIP: Allow rolling out PR status reporting

wfleming:
Adds new config values to allow reporting PR statuses for only specific
users and/or a percentage of users.

The `author_username` key being used in the payload isn't actually sent
right now: I need to evaluate if there's an existing key we send that is
usable, or if we need to thread this data through on the app side.

cc @codeclimate/review

### DIFF
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -13,6 +13,14 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
       label: "Github Context",
       description: "The integration name next to the pull request status",
       default: "codeclimate"
+    attribute :rollout_github_usernames, Axiom::Types::String,
+      label: "Allowed Author's Usernames",
+      description: "The GitHub usernames of authors to report status for, comma separated",
+      default: nil
+    attribute :rollout_percentage, Axiom::Types::Integer,
+      label: "Author Rollout Percentage",
+      description: "The percentage of users to report status for",
+      default: nil
 
     validates :oauth_token, presence: true
   end
@@ -21,6 +29,28 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
   self.description = "Update pull requests on GitHub"
 
   private
+
+  def report_status?
+    if author_username.present? && (config.rollout_github_usernames.present? || config.rollout_percentage.present?)
+      rollout_allowed_by_username? || rollout_allowed_by_percentage?
+    else
+      true
+    end
+  end
+
+  def rollout_allowed_by_username?
+    config.rollout_github_usernames.present? &&
+      config.rollout_github_usernames.split(",").map(&:strip).include?(author_username)
+  end
+
+  def rollout_allowed_by_percentage?
+    config.rollout_percentage.present? &&
+      Digest::MD5.hexdigest(author_username).to_i(16) % 100 <= config.rollout_percentage
+  end
+
+  def author_username
+    @payload["author_username"]
+  end
 
   def update_status_skipped
     update_status("success", presenter.skipped_message)
--- a/lib/cc/services/github_pull_requests.rb
+++ b/lib/cc/services/github_pull_requests.rb
@@ -13,7 +13,7 @@ class CC::Service::GitHubPullRequests < CC::PullRequests
       label: "Github Context",
       description: "The integration name next to the pull request status",
       default: "codeclimate"
-    attribute :rollout_github_usernames, Axiom::Types::String,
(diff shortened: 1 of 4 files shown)


wfleming:
The CC issue is because this class finally got to be too long. I would like to ignore that.

lib/cc/services/github_pull_requests.rb
```diff
@@ -22,6 +30,37 @@ class Config < CC::Service::Config
...
+    if should_apply_rollout?
+      rollout_allowed_by_username? || rollout_allowed_by_percentage?
+    else
+      true
+    end
+  end
+
+  def should_apply_rollout?
+    (github_login.present? || github_user_id.present?) &&
+      (config.rollout_usernames.present? || config.rollout_percentage.present?)
```
pbrisbin: I think these has some spots in the boolean grid that wouldn't work (login present, but rollout-usernames not present). Should it be this:

```rb
(github_login.present? && config.rollout_usernames.present?) ||
  (github_user_id.present? && config.rollout_percentage.present?)
```

?

lib/cc/services/github_pull_requests.rb
```diff
@@ -22,6 +30,37 @@ class Config < CC::Service::Config
...
+  end
+
+  def rollout_allowed_by_username?
+    github_login.present? && config.rollout_usernames.present? &&
+      config.rollout_usernames.split(",").map(&:strip).include?(github_login)
+  end
+
+  def rollout_allowed_by_percentage?
+    github_user_id.present? && config.rollout_percentage.present? &&
+      github_user_id % 100 <= config.rollout_percentage
```
pbrisbin: I _think_ `github_user_id` is a string

lib/cc/services/github_pull_requests.rb
```diff
@@ -22,6 +30,37 @@ class Config < CC::Service::Config
...
+  end
+
+  def rollout_allowed_by_username?
+    github_login.present? && config.rollout_usernames.present? &&
+      config.rollout_usernames.split(",").map(&:strip).include?(github_login)
+  end
+
+  def rollout_allowed_by_percentage?
+    github_user_id.present? && config.rollout_percentage.present? &&
+      github_user_id % 100 < config.rollout_percentage
```
pbrisbin: Comment got hidden in the diff: I _think_ `github_user_id` is a string.

pbrisbin:
Two minor thoughts, but LGTM besides. We should be careful that we don't install this in app until the UI is correctly hiding the new attributes.

lib/cc/services/github_pull_requests.rb
wfleming: It's an `Integer` field on `PullRequest`. Since we're pulling it from there for the payloads, I think we'll still get it as an int here?

lib/cc/services/github_pull_requests.rb
wfleming: Yeah, that does make sense, I'll make that change.

lib/cc/services/github_pull_requests.rb
pbrisbin: Just confirmed it is an `Integer` in the db. I'd also assume that's what you'll get here too.

wfleming:
> We should be careful that we don't install this in app until the UI is correctly hiding the new attributes.

The [app form](https://github.com/codeclimate/app/blob/master/app/views/services/_github_pull_requests_form.html.slim) doesn't auto-generate anything from config, so I don't believe that would be a concern. (Also it looks like Devon's PR shipped.)

pbrisbin:
Nice. Thanks for checking that.

wfleming:
Ready for re-review @pbrisbin

lib/cc/services/github_pull_requests.rb
```diff
@@ -39,8 +39,8 @@ def report_status?
   end
 
   def should_apply_rollout?
-    (github_login.present? || github_user_id.present?) &&
-      (config.rollout_usernames.present? || config.rollout_percentage.present?)
+    (github_login.present? || config.rollout_usernames.present?) &&
+      (github_user_id.present? || config.rollout_percentage.present?)
```
dblandin: I think the conditional operators need to change here as well.

(github_login.present? && config.rollout_usernames.present?) ||
  (github_user_id.present? && config.rollout_percentage.present?)

wfleming: Good catch, thanks, addressed.

pbrisbin:
LGTM

PR #118, opened 2016-09-21
Add SafeWebhook concept

pbrisbin:
Implements an IP address blacklist to mitigate users configuring a Webhook to
POST to one of our (presumably unauthenticated) internal services.

http://blog.fanout.io/2014/01/27/how-to-safely-invoke-webhooks/#ip-address-blacklisting

- Resolve the configured value to its canonical IP address
- Compare that against the blacklist
- Rewrite the request to be for that verified IP with an explicit Host header

This is disable-able via ENV since CC:E will be posting to local services
intentionally.

/cc @codeclimate/review @brynary

TODO:

- [x] Finalize the blacklist, i.e. should we add `192.` or anything else?
- [x] Is it possible certain systems would return IPv6 from `Resolv::DNS`?

### DIFF
--- a/lib/cc/service/http.rb
+++ b/lib/cc/service/http.rb
@@ -1,5 +1,6 @@
 require "active_support/concern"
 require "cc/service/response_check"
+require "cc/service/safe_webhook"
 
 module CC::Service::HTTP
   extend ActiveSupport::Concern
@@ -38,10 +39,8 @@ module CC::Service::HTTP
   end
 
   def raw_get(url = nil, params = nil, headers = nil)
-    http.get do |req|
-      req.url(url) if url
+    http_method(:get, url, nil, headers) do |req|
       req.params.update(params) if params
-      req.headers.update(headers) if headers
       yield req if block_given?
     end
   end
@@ -59,6 +58,11 @@ module CC::Service::HTTP
       req.headers.update(headers) if headers
       req.body = body if body
       block.call req if block
+
+      unless allow_internal_webhooks?
+        safe_webhook = CC::Service::SafeWebhook.new(url)
+        safe_webhook.validate!(req)
+      end
     end
   end
 
@@ -98,5 +102,10 @@ module CC::Service::HTTP
       status: response.status,
       message: "Success",
     }
+  end
+
+  def allow_internal_webhooks?
+    var = ENV["CODECLIMATE_ALLOW_INTERNAL_WEBHOOKS"] || ""
+    var == "1" || var == "true"
   end
 end
--- a/lib/cc/service/safe_webhook.rb
+++ b/lib/cc/service/safe_webhook.rb
@@ -1,0 +1,71 @@
+require "ipaddr"
+require "resolv"
+
+module CC
+  class Service
+    class SafeWebhook
+      InvalidWebhookURL = Class.new(StandardError)
+
+      # https://en.wikipedia.org/wiki/Private_network#Private_IPv4_address_spaces
+      # https://en.wikipedia.org/wiki/Private_network#Private_IPv6_addresses
+      PRIVATE_ADDRESS_SUBNETS = [
+        IPAddr.new("10.0.0.0/8"),
+        IPAddr.new("172.16.0.0/12"),
+        IPAddr.new("192.168.0.0/16"),
+        IPAddr.new("fd00::/8"),
+        IPAddr.new("127.0.0.1"),
+        IPAddr.new("0:0:0:0:0:0:0:1"),
+      ].freeze
+
+      def self.getaddress(host)
+        @resolv ||= Resolv::DNS.new
+        @resolv.getaddress(host).to_s
+      end
+
+      def initialize(url)
+        @url = url
+      end
+
+      # Resolve the Host to an IP address, validate that it doesn't point to
+      # anything internal, then alter the request to be for the IP directly with
+      # an explicit Host header given.
+      #
+      # See http://blog.fanout.io/2014/01/27/how-to-safely-invoke-webhooks/#ip-address-blacklisting
+      def validate!(request)
+        uri = URI.parse(url)
+        address = self.class.getaddress(uri.host)
+
+        if internal?(address)
+          raise_invalid("resolves to a private IP address")
+        end
+
+        alter_request(request, uri, address)
+      rescue URI::InvalidURIError, Resolv::ResolvError, Resolv::ResolvTimeout => ex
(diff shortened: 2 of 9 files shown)


lib/cc/service/http.rb
```diff
@@ -99,4 +105,9 @@ def formatted_post_response(response, url, body)
       message: "Success",
     }
   end
+
+  def skip_webhook_validation?
+    var = ENV["CODECLIMATE_SKIP_WEBHOOK_VALIDATION"] || ""
```
wfleming: We probably should have done this for the pr rollout too. Oh well.

lib/cc/service/safe_webhook.rb
```diff
@@ -0,0 +1,54 @@
...
+      rescue URI::InvalidURIError, Resolv::ResolvError => ex
+        raise_invalid(ex.message)
+      end
+
+      private
+
+      attr_reader :url
+
+      def internal?(address)
+        address.start_with?("127.") ||
```
wfleming: We should probably also filter `192.168` & a few others to be cautious?

https://en.wikipedia.org/wiki/Reserved_IP_addresses

lib/cc/service/safe_webhook.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def alter_request(request, uri, address)
+        address_uri = uri.dup
+        address_uri.host = address
+        request.url(address_uri.to_s)
+        request.headers.update(Host: uri.host)
+      end
+
+      def raise_invalid(message)
+        raise InvalidWebhookURL, "The Webhook URL #{url} is invalid: #{message}"
```
wfleming: I'm a bit surprised we're just raising this, and not using the catch above to add a validation error of some kind. Are you planning to catch this app-side & do something with it or just let it blow up?

Rakefile
```diff
@@ -1,8 +1,6 @@
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 
-RSpec::Core::RakeTask.new(:spec) do |task|
-  task.ruby_opts = "-W"
```
wfleming: @maxjacobson is probably gonna nit you for this 😉. The warning flag previously helped find some pretty old Ruby constructs. Maybe we don't need it anymore though it there are no current warnings? (We're probably unlikely to introduce new ones.)

lib/cc/service/safe_webhook.rb
pbrisbin: Yeah, it's a TODO in the PR description to finalizer this. that wikipedia page or some RFC should give us something comprehensive to use.

Rakefile
pbrisbin: I only removed it because of the **incredibly stupid omg rubby why you so dumb** warning about private `attr_reader`s (one of which I use in `SafeWebhook`.

Is there a better way to handle that philosophical disagreement I have with rubby?

I get that this codebase had some cruft that was found via `-W`, but we don't run with warnings on typically, and now that this codebase was  brought inline, maybe it's no longer needed?

maxjacobson: IIRC I actually removed it, you defended its presence, and then later converted the code base to use rspec and removed it at which point I proposed keeping it... kind of a long and complicated relationship.... anyway I'm OK with turning it off because most of the compiler warnings are also rubocop cops which lets us display them in cc which is nice. The only reason I would want to keep them on at this point is because this is a gem and gems shouldn't generate warnings in case their consumers have warnings turned on. But I think this is only a gem for convenience, so it's less important. Also that warning is annoying.

pbrisbin:
So... the two TODOs I noted in the PR description :) I'll let you know when addressed.

brynary:
tl;dr yes :)

pbrisbin:
@brynary I've addressed the two TODOs to the best of my ability, would appreciate your review now.

- Added reserved private IPv4 prefixes according to RFC
- Added reserved private IPv6 prefix according to wikipedia

Some caveats:

I'm including localhost prefixes at the moment. `Resolv::DNS` raises an error when trying to resolve `"localhost"`, so I don't think they're strictly needed. WDYT?

I am unable to actually get `Resolv::DNS` to return an `IPv6` address, even through the documentation says it may. I've tried calling `#each_address` on domains known to be using IPv6 and I always only get back IPv4 addresses. Therefore, I'm less confident in that blacklist entry as its based on the assumptions that I'm interpreting the wikipedia page correctly and `Resolv::DNS` _would_ represent such an address with that prefix.

> IIRC, there's a handy object somewhere in Ruby stdlib which can be used to check an IP address against a subnet mask

I'm going to look into this now, but I think the prefix checking is equivalent. It's possible moving to that library function will require IPv4-vs-IPv6 branching, which I'd like to avoid.

Question:

I understand this will often be disabled in CCE. Should I set that up as:

- Hard-coded to not validate, not configurable
- Configurable, with the default being to validate
- Configurable, with the default being not to validate

brynary:
codeclimate reported an errored status here... Also, I noticed an addition of a `.codeclimate.yml`... Can this be split out?

lib/cc/service/http.rb
brynary: Would it make more sense to call this something like `ALLOW_INTERNAL_WEBHOOKS`? "Validation" seems ambiguous.

brynary:
I thought I saw somewhere that this only applies to non-GET verbs... Is that accurate? If so, is there any reason not to apply this for all HTTP verbs?

pbrisbin:
> I thought I saw somewhere that this only applies to non-GET verbs... Is that accurate? If so, is there any reason not to apply this for all HTTP verbs?

Yes that's accurate. For whatever reason, this codebase has GET requests going through a different codepath than non-GET. I didn't find it valuable to get into the weeds on that since this protection is not needed on GET requests. I can do it now though, as I sense you disagree.

pbrisbin:
> codeclimate reported an errored status here

bundler-audit errored because we gitignore the Gemfile.lock in here -- no idea why.

>  Also, I noticed an addition of a .codeclimate.yml... Can this be split out

I don't see the value in doing that, but sure.

lib/cc/service/http.rb
pbrisbin: Done

lib/cc/service/safe_webhook.rb
pbrisbin: Ignoring ;)

pbrisbin:
@brynary ready for another look!

PR #120, opened 2016-09-26
Don't run specs with warnings enabled

pbrisbin:
On balance, these warnings aren't valuable. They mostly generate noise for
things we actually prefer and don't consider problematic (e.g. private
attributes). The warnings we do care about will be raised as Rubocop issues.

/cc @codeclimate/review @wfleming @maxjacobson


maxjacobson:
I'm fine with this

wfleming:
sure sure

PR #121, opened 2016-09-27
Revert "Add SafeWebhook concept"

pbrisbin:
Reverts codeclimate/codeclimate-services#118

This approach does not work because rewriting the request to be for the IP address fails SSL host verification (which does not use the Host header).


wfleming:
Oops. LGTM.

Is the fix going to be to verify the IP but not re-write the request?

pbrisbin:
I think that's a non-starter. If you don't re-write the request you can't be sure the resolve you just did is the same answer as the one that ultimately leads to the connection -- the blog post that outlined this mitigation was pretty adamant about that.

Then again, maybe that's better than nothing? /cc @brynary

wfleming:
Yeah, there's a race where a malicious DNS server could change its answer.

It might be possible to do this with Net::HTTP directly: open the connect, inspect the connection object or use `#trace` to get the server IP, validate that, and then do the post request within the same connection scope. I think that would prevent the malicious DNS issue?

PR #122, opened 2016-09-28
SafeWebhook - Take 2

pbrisbin:
/cc @codeclimate/review

Found a way to do the SafeWebhook thing that doesn't require re-writing the request,
but does ensure the Resolv done when checking if internal gives the same answer
when we ultimately connect.

TODO:

- [x] Fix the service tests that make host-less requests
- [x] Consider how to avoid actual DNS resolution in specs (they're slow now)
- [x] Test `SafeWebhook`, itself and users of it
- [x] Avoid require side-effect
- [x] Raise meaningful exception

### DIFF
--- a/lib/cc/service/http.rb
+++ b/lib/cc/service/http.rb
@@ -1,5 +1,6 @@
 require "active_support/concern"
 require "cc/service/response_check"
+require "cc/service/safe_webhook"
 
 module CC::Service::HTTP
   extend ActiveSupport::Concern
@@ -38,10 +39,8 @@ module CC::Service::HTTP
   end
 
   def raw_get(url = nil, params = nil, headers = nil)
-    http.get do |req|
-      req.url(url) if url
+    http_method(:get, url, nil, headers) do |req|
       req.params.update(params) if params
-      req.headers.update(headers) if headers
       yield req if block_given?
     end
   end
@@ -53,6 +52,8 @@ module CC::Service::HTTP
 
   def http_method(method, url = nil, body = nil, headers = nil)
     block = Proc.new if block_given?
+
+    CC::Service::SafeWebhook.ensure_safe!(url)
 
     http.send(method) do |req|
       req.url(url) if url
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,6 +9,9 @@ engines:
     enabled: true
   rubocop:
     enabled: true
+    exclude_fingerprints:
+      # Using #=== intentionally to do subnet masking
+      - d1afe90be49c43e85a76bfa58f637804
 ratings:
   paths:
   - "**.rb"
--- a/lib/cc/fixed_resolv.rb
+++ b/lib/cc/fixed_resolv.rb
@@ -1,0 +1,29 @@
+require "resolv-replace"
+
+module CC
+  class FixedResolv < Resolv::DNS
+    def self.enable!
+      new.tap do |instance|
+        Resolv::DefaultResolver.replace_resolvers([instance])
+      end
+    end
+
+    def initialize
+      @addresses = {}
+    end
+
+    def setaddress(name, address)
+      addresses[name] = address
+    end
+
+    def each_address(name)
+      if addresses.key?(name)
+        yield addresses.fetch(name)
+      end
+    end
+
+    private
+
+    attr_reader :addresses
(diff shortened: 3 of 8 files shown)


lib/cc/service/http.rb
```diff
@@ -59,6 +58,10 @@ def http_method(method, url = nil, body = nil, headers = nil)
       req.headers.update(headers) if headers
       req.body = body if body
       block.call req if block
+
+      unless CC::Service::SafeWebhook.safe?(url)
+        raise "Invalid Webhook"
```
wfleming: Seems like we should have a class for this exception and not use `StandardError`?

pbrisbin: Sorry, meant to include that in my TODOs

lib/cc/resolv_override.rb
```diff
@@ -0,0 +1,46 @@
...
+      private
+
+      def addresses
+        @addresses ||= {}
+      end
+    end
+  end
+end
+
+Resolv::DefaultResolver.replace_resolvers([
```
brynary: Require side effect here... Interesting potential CC check :)

lib/cc/resolv_override.rb
```diff
@@ -0,0 +1,46 @@
...
+  CC::ResolvOverride.fixed,
+  CC::ResolvOverride.dns,
```
brynary: Do we still want DNS at all?

lib/cc/service/http.rb
brynary: FYI, a bare raise is RuntimeError not StandardError 🤓

lib/cc/resolv_override.rb
pbrisbin: At present, doing nothing else, the resolv-override is going to be active in any app container (since it loads the cc-services code, which loads this code). So if we dropped `dns` here, other calls (like App hitting GH API or something) would also go by way of `fixed` which (currently) doesn't fall back if you haven't pre-set the host->address case.

So that would all have to be addressed somehow; then there might be other interactions I'm _not_ thinking about. Ultimately, I'm not sure the of the upside.

lib/cc/resolv_override.rb
pbrisbin: Intentional. This seems to be the simplest way to achieve what I wanted, do you have any concrete downsides or suggested alternative approaches to where/how to actually trigger the `replace_`?

brynary: Downside would be that if this gem is added to any project (and it's OSS, so possible), and is required (either explicitly or via auto-require) that this behavior may take effect? (Even if someone is just adding this gem to e.g. use one specific behavior)

Safer would be to expose a class method like `CC::Services.override_dns_resolution` or similar and then call that from _outside_ the gem if you want that behavior.

lib/cc/resolv_override.rb
brynary: See the comment I just made about `CC::Services.override_dns_resolution`... If we do that, and we only call that method from the containers which are for sending events, then we have the benefit of not affecting the resolver logic in non-services containers and fixing it completely in services containers.

lib/cc/resolv_override.rb
pbrisbin: Works for me.

pbrisbin:
I have some new concerns about this approach:

In general the idea is to resolve the host to an address, check that that address is not internal, and ensure that the future resolution to actually connect to it gets the same answer.

The solution here is essentially caching. We override the default resolution behavior of the system and make it so that when the eventual connection resolves, it gets the same answer we cached earlier as part of checking it.

Now we have a cache invalidation problem -- what happens when GitHub updates its DNS?

I'm assuming I need to implement some sort of TTL, and keep it lower than DNS TTL. Am I missing anything easier?

/cc @wfleming @brynary

pbrisbin:
I guess one way that would work is:

- Actually resolv dns
- Check not internal
- Add override for future resolvs (even if already overridden)
- Proceed

That should avoid stale entries being used, provided we only add overrides such that we'll _always_ add new overrides before later connections are attempted.

I'll go down that route...

pbrisbin:
Updated and rebased. This is ready for more final review.

wfleming:
LGTM

PR #123, opened 2016-10-04
Configure new CC test reporter

pbrisbin:
/cc @codeclimate/review

This was extracted from #122 where the DNS stubbing we're doing breaks the
integrated test reporter.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,3 +4,8 @@ source "https://rubygems.org"
 gemspec
 
 gem "pry"
+
+group :test do
+  gem "simplecov"
+  gem "codeclimate-test-reporter", "1.0.0.pre.rc2"
+end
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -4,3 +4,8 @@ source "https://rubygems.org"
 gemspec
 
 gem "pry"
+
+group :test do
+  gem "simplecov"
+  gem "codeclimate-test-reporter", "1.0.0.pre.rc2"
```
dblandin: Is there an important distinction between keeping these in the gemspec as development dependencies and moving them here to a test group?

pbrisbin: No idea. Doing it this way was easier copy/pasting from the instructions.

dblandin: Ok :)

dblandin:
LGTM

PR #124, opened 2016-11-01
Make SafeWebhook safe again

pbrisbin:
Since we still do invoke service code from non-cc-services codebases, leaving
the default resolver as overridden will cause problems.

This patch changes the approach to only override DNS for the duration of a
block. Additionally, we de-coupled the DNS override from the safe-webhook
check. It means both things need to happen to be fully safe, but each object
and be more singly-responsible, which helps understanding and testing.

/cc @codeclimate/review


wfleming:
Gonna nit you: this approach still isn't thread-safe, is it? Any thread executing in parallel with the block would also get the overidden resolver? I don't think that would be a problem for our usage, just want to understand the boundaries.

pbrisbin:
That sounds right, but I think it's OK:

- Block opens, default resolvers replaced
- *Any other threads would still get a real DNS response here*
- This thread makes an actual DNS request via `SafeWebhook`
- *Any other threads would get a "fixed", but accurate response here*
- This thread gets a fixed DNS response for the actual HTTP call
- Block closes, default resolvers restored

Since there's no way to just ask for the default resolvers (which I was annoyed by before, but now it works in my favor) there's no danger of the other threads hanging onto a `CC::Resolv::Fixed` and using it later, which would be the only dangerous scenario.

Am I missing anything?

pbrisbin:
To clarify the invariant a bit more:

- While the block is open, all DNS queries for any threads will get the same answer every time
- The responses to those queries may be "real" or "fixed", but they'll be as fresh as the time since the block was opened
- The block won't be open for very long

If all those are true, I think everyone's happy?

wfleming:
Yeah, I think that's correct, and I agree it's safe for our needs.

wfleming:
LGTM

PR #126, opened 2016-11-18
Shorten test coverage status

maxjacobson:
Our messages are being truncated to, e.g. "58.02% test coverage (+1..."

The status already has a name of "codeclimate/coverage", so I think with
context this will be sufficient information.

Another idea considered: "58.02% cov (+1%)" or "58.02% (+1%) cov". We use similar language in the extension ("100% cov") so maybe that's good for consistency. Thoughts welcome!

@codeclimate/copy @codeclimate/review

### DIFF
--- a/lib/cc/presenters/pull_requests_presenter.rb
+++ b/lib/cc/presenters/pull_requests_presenter.rb
@@ -28,7 +28,7 @@ module CC
       end
 
       def coverage_message
-        message = "#{formatted_percent(@covered_percent)}% test coverage"
+        message = "#{formatted_percent(@covered_percent)}%"
 
         if @covered_percent_delta > 0
           message += " (+#{formatted_percent(@covered_percent_delta)}%)"
--- a/spec/cc/presenters/pull_requests_presenter_spec.rb
+++ b/spec/cc/presenters/pull_requests_presenter_spec.rb
@@ -22,15 +22,15 @@ describe CC::Service::PullRequestsPresenter, type: :service do
   end
 
   it "message coverage same" do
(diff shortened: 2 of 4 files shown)


wfleming:
I'm in favor of this (I'd also be in favor of similar changes to the issues status), since GH truncates so much in their UI, but this should probably get sign of from @brynary.

maxjacobson:
Yep. The issues status for this PR currently reads "Code Climate didn't find any new or f..." 😞 

There's no character length limitation -- the truncation is applied with CSS based on a width percentage... but I'm sure we can find a way to express what we need to express more succinctly.

GordonDiggs:
It seems excessive - have we thought about reaching out to GH to see if they'll widen it?

maxjacobson:
Not that I'm aware of, but yea, doesn't hurt to ask the question :)

GordonDiggs:
👍 seems worthwhile. Still good to ship this PR too IMO

maxjacobson:
Done. I'll let you know if I hear anything. It will be helpful for making a decision about the issues status check, which is not changed by this PR.

WDYT @brynary?

maxjacobson:
Heard back from GitHub. The support person I talked with agreed the truncation "could be improved" and opened an internal issue about it with their design team, but gave no time frame.

Also chatted IRL with @brynary and agreed on the change done in this PR, so I'm going to ship this change for coverage. We talked about another potential change for our issues status check to make that shorter, and I think I'll just capture that in our JIRA for someone else to pick up. Maybe the CSS will be changed by the time someone does and it won't be necessary. IMO it's lower priority. The coverage status contains meaningful information (the delta) that (currently) can't be seen anywhere else. The issues are visible by clicking through, or via the extension.

PR #130, opened 2016-12-16
Support authorizing asana requests via tokens

maxjacobson:
Asana is deprecating their API Key access, as reported by a helpful
user.

https://asana.com/developers/documentation/getting-started/auth#api-key

We confirmed with their support team that they're planning to drop
support for API Key authorization in "early 2017".

This PR introduces support for authorizing with personal access tokens,
and attempts to nudge users in that direction gently. I think we'll need
to make a plan about how to migrate existing repos over.

### DIFF
--- a/lib/cc/services/asana.rb
+++ b/lib/cc/services/asana.rb
@@ -1,6 +1,7 @@
 class CC::Service::Asana < CC::Service
   class Config < CC::Service::Config
-    attribute :api_key, Axiom::Types::String, label: "API key"
+    attribute :personal_access_token, Axiom::Types::String, label: "Personal Access Token"
+    attribute :api_key, Axiom::Types::String, label: "API key (Deprecated)"
 
     attribute :workspace_id, Axiom::Types::String, label: "Workspace ID"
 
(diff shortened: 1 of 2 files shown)


lib/cc/services/asana.rb
```diff
@@ -1,6 +1,7 @@
 class CC::Service::Asana < CC::Service
   class Config < CC::Service::Config
-    attribute :api_key, Axiom::Types::String, label: "API key"
+    attribute :personal_access_token, Axiom::Types::String, label: "Personal Access Token"
```
GordonDiggs: Can we treat this like the GH tokens and only display the full thing once?

maxjacobson: Sure, good call. I'll open an app PR that uses this branch and adds the obfuscation there

GordonDiggs: Cool - wasn't sure if we needed to do something special here as well

maxjacobson: I thought we would need to do something special, but while testing locally it seems like everything Just Works ™️ even without changing the views in app :)

GordonDiggs: Even better!

PR #132, opened 2016-12-19
Change success message when approved_by present

ABaldwinHunter:
@codeclimate/review 

Supports https://github.com/codeclimate/api/pull/149

I considered letting the client construct and pass the message, but thought this approach was a little more restricting in a good way. Thoughts welcome

I also considered, initially, passing the approval_id, but I don't think we want this gem to know about all of our data types.

### DIFF
--- a/lib/cc/presenters/pull_requests_presenter.rb
+++ b/lib/cc/presenters/pull_requests_presenter.rb
@@ -13,6 +13,12 @@ module CC
 
         @covered_percent = payload["covered_percent"]
         @covered_percent_delta = payload["covered_percent_delta"]
+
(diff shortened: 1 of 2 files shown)


lib/cc/presenters/pull_requests_presenter.rb
```diff
@@ -40,7 +46,9 @@ def coverage_message
       end
 
       def success_message
-        if @new_count > 0 && @fixed_count > 0
+        if @approved_by
```
GordonDiggs: probably good to add `.present?` here to avoid a weird situation if it's an empty string

ABaldwinHunter:
@GordonDiggs ready for another look!

PR #135, opened 2017-05-02
Loosen nokogiri dependency

pbrisbin:
In order to address
https://github.com/sparklemotion/nokogiri/issues/1615, we need to
upgrade to nokogiri-1.7.1. Applications also depending on this gem are
in conflict unless this version constraint is loosened.


maxjacobson:
The change LGTM, but the tests are failing now. Can take another look if the changes to make it green are meaningful/risky

pbrisbin:
Oh man, I seem to have stepped in it. First theory is the loosened nokogiri allowed a different faraday that breaks... Will investigate.

pbrisbin:
#136

maxjacobson:
LGTM on green

PR #137, opened 2017-06-29
Add Token marker type

maxjacobson:
This will make it simpler for us to identify which fields are sensitive
by type, rather than by field name.


maxjacobson:
makes sense, good catch. I didn't realize we weren't using the gem in that repo. I think I see what you mean and will ping you there for review

PR #141, opened 2018-06-28
Update to Ruby 2.4.4 and set homepage URL.

wilson:


codeclimate-services.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
   spec.email = ["bryan@brynary.com"]
   spec.summary = "Service classes for Code Climate"
   spec.description = "Service classes for Code Climate"
-  spec.homepage = ""
+  spec.homepage = "https://developer.codeclimate.com/"
```
wfleming: developer.codeclimate.com is (perhaps confusingly) dedicated to API docs, which aren't connected to this gem, so this might not be the right choice. I don't know if this is better or worse than plain codeclimate.com.

PR #145, opened 2021-02-25
QUA-179: Update Jira integration description

noelia-lencina:
https://codeclimate.atlassian.net/browse/QUA-179

### DIFF
--- a/lib/cc/services/jira.rb
+++ b/lib/cc/services/jira.rb
@@ -3,7 +3,7 @@ require "base64"
 class CC::Service::Jira < CC::Service
   class Config < CC::Service::Config
     attribute :domain, Axiom::Types::String,
-      description: "Your JIRA host domain (e.g. yourjira.com[:PORT], please exclude https://. Include PORT only if you are using a port number to access your own server)"
+      description: "Your JIRA host domain (e.g. yourjira.com[:PORT], please exclude https:// and only add PORT if you are using a different port number to access your own server that is not the default)"
 
     attribute :username, Axiom::Types::String,
       description: "Must exactly match the 'username' that appears on your JIRA profile page."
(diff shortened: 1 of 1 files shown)


noelia-lencina:
CI error: I think it's because we are using the old configuration for Circle CI. I will fix it in another PR and rebase this.

lib/cc/services/jira.rb
```diff
@@ -3,7 +3,7 @@
 class CC::Service::Jira < CC::Service
   class Config < CC::Service::Config
     attribute :domain, Axiom::Types::String,
-      description: "Your JIRA host domain (e.g. yourjira.com:PORT, please exclude https://)"
+      description: "Your JIRA host domain (e.g. yourjira.com[:PORT], please exclude https://. Include PORT only if you are using a port number to access your own server)"
```
filipesperandio, line 6:
```suggestion
      description: "Your JIRA host domain (e.g. yourjira.com[:PORT], please exclude "https://" and only  add PORT if you are using a different port number to access your own server that is not the default)"
```

filipesperandio (approved):

lib/cc/services/jira.rb
noelia-lencina: I think that even if you are using the default port, you need to include the port number here. You don't need the port number if you are not hosting your own Jira instance or if you are but have a DNS configured so you can access the instance just by its domain.

fede-moya (approved):

lib/cc/services/jira.rb
filipesperandio: PORT follows its protocol, for https it is 443 - it is the only use I see, so if not specified, 443 is what it is gonna be, right?

filipesperandio: feel free to drop it if  it made things more confusing.  🤦

filipesperandio:
Seems like we need to port this over to circle 2... too long we don't do anything here.

PR #146, opened 2021-02-25
Port to Circle CI 2

noelia-lencina:


fede-moya (approved):
should we remove the old circle config ?, just curious do you know how this services get deployed or something ?

filipesperandio (approved):

PR #150, opened 2022-06-01
Support ruby 2.7

giordanoluzardo:
## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Description

Starting on Ruby 2.7  executing `nil.to_s` produces a `Frozen String` error. It's documented explicitly in the [release notes](https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/).

> 
>Module#name, true.to_s, false.to_s, and nil.to_s now always return a frozen String. The returned String is always the same >for a given object. [Experimental] [[Feature #16150](https://bugs.ruby-lang.org/issues/16150)] 
>

As far as I can tell this Ruby change it's the only thing that blocks this gem from supporting Ruby 2.7. 

Adapting the code so it avoids running `nil.to_s`


fede-moya:
@giordanoluzardo according to;
https://github.com/codeclimate/codeclimate-services/blob/0caac6294caef123b16d1f6e3d6d4bd0c26cf95f/lib/cc/service/formatter.rb#L22

`nil.to_s` will never be executed but `"".to_s` will be doesn't it ?

fede-moya:
> @giordanoluzardo according to;
> 
> https://github.com/codeclimate/codeclimate-services/blob/0caac6294caef123b16d1f6e3d6d4bd0c26cf95f/lib/cc/service/formatter.rb#L22
> 
> `nil.to_s` will never be executed but `"".to_s` will be doesn't it ?

**Update**

I believe I reasoned in the same wrongly way as the one who wrote this code. I briefly assumed that if sending `fetch` with an optional value to hash that has the key set to nil it would return the default but according to the docs it only returns the default value if the key is not present.
https://apidock.com/ruby/Hash/fetch


In any case, it's clear that the intention was that if there is no given prefix set it to "". So, indeed, I believe the error that we are seeing in app makes sense and it's worth being patched. Well done 🚀

lib/cc/service/formatter.rb
```diff
@@ -19,7 +19,7 @@ def service_title
   end
 
   def message_prefix
-    prefix = options.fetch(:prefix, "").to_s
+    prefix = options.fetch(:prefix, "").to_s.dup
```
fede-moya, line 22: I believe given the previously shared reasoning this is more appropriate 
```suggestion
    prefix = (options[:prefix] || "").to_s
```

fede-moya (commented):
Left a minor suggestion, wdyt ?

Gemfile
```diff
@@ -7,5 +7,4 @@ gem "pry"
 
 group :test do
   gem "simplecov"
-  gem "codeclimate-test-reporter", "~> 1.0.0"
```
fede-moya, line 10: 👍🏼

lib/cc/service/formatter.rb
giordanoluzardo: Fede, I remember trying this option too and both work as well, I'll take your suggestion 🙏

.circleci/config.yml
```diff
@@ -1,15 +1,20 @@
...
+version: 2.0
 jobs:
   test:
+    environment:
+      CC_TEST_REPORTER_ID: ad21bc049a323138b0a88086069115e9df6dba417912095b7bb17b7a65fc6182
     docker:
-      - image: circleci/ruby:2.4.4
-    setup_remote_docker:
-      docker_layer_chaching: true
+      - image: "circleci/ruby:2.4.4"
```
fede-moya, line 7: ?
```suggestion
      - image: "circleci/ruby:2.7.5"
```

fede-moya:
I guess we should consider bumping `.ruby_version` as well

fede-moya (approved):
Nicely done 👍🏼

larkinscott (approved):
LGTM!

PR #152, opened 2022-06-29
QUA-360: Add skip diff coverage method

camillof:
### Context

Now that we're implementing the ability to skip draft PRs reviews, we want to avoid the "waiting for status to be reported" situation on the diff and total coverage statuses, specifically when you configured them as a protection branch rule.

### Changes

- Introduce a new method to report a skipped diff/total coverage status
- Done some refactoring (which can be ignored if we don't like it)
- Added tests for the `CC::PullRequests` class

### Related issue

https://codeclimate.atlassian.net/browse/QUA-360

### DIFF
--- a/lib/cc/pull_requests.rb
+++ b/lib/cc/pull_requests.rb
@@ -31,6 +31,19 @@ class CC::PullRequests < CC::Service
     response
   end
 
+  def receive_pull_request_diff_coverage
+    setup_http
+    state = @payload["state"]
+
+    if state == "skipped" && report_status?
+      update_diff_coverage_status_skipped
+    else
+      @response = simple_failure("Unknown state")
+    end
+
+    response
+  end
+
   private
 
   def report_status?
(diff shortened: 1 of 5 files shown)


lib/cc/pull_requests.rb
```diff
@@ -6,29 +6,15 @@ def receive_test
...
-    state = @payload["state"]
-
-    if %w[pending success failure skipped error].include?(state) && report_status?
-      send("update_status_#{state}")
-    else
-      @response = simple_failure("Unknown state")
-    end
-
-    response
+    receive_request(%w[pending success failure skipped error], :update_status)
```
dantevvp, line 9: Is there a reason to pass the `call_method` as a symbol instead of a string? I see that we convert it as a string later in the `receive_request` method

dantevvp (approved):
Looks good! Just one small doubt

lib/cc/pull_requests.rb
camillof: Honestly, there isn't a strong reason. I just felt like it was more readable, but I'm open to change it if you don't feel the same about it

dantevvp: I personally like it as a symbol. I associate symbols with methods better :+1:

PR #154, opened 2022-07-05
QUA-360: Separate diff and total coverage in different methods

camillof:
### Context

Following #152 where we added a single method for skipping `diff` and `total` coverage, we realized that we may need them separated, to be called independently.

In addition to the `skipped` method, we also need the ability to set the `pending` status for bot total and diff coverage.

image

### DIFF
--- a/lib/cc/pull_requests.rb
+++ b/lib/cc/pull_requests.rb
@@ -14,7 +14,11 @@ class CC::PullRequests < CC::Service
   end
 
   def receive_pull_request_diff_coverage
-    receive_request("skipped", :update_diff_coverage_status)
+    receive_request(%w[pending skipped], :update_diff_coverage_status)
+  end
+
+  def receive_pull_request_total_coverage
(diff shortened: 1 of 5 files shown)


f-moya (approved):

lib/cc/pull_requests.rb
```diff
@@ -14,7 +14,11 @@ def receive_pull_request_coverage
   end
 
   def receive_pull_request_diff_coverage
-    receive_request("skipped", :update_diff_coverage_status)
+    receive_request(%w[pending skipped], :update_diff_coverage_status)
```
giordanoluzardo, line 17: 👍

lib/cc/pull_requests.rb
```diff
@@ -14,7 +14,11 @@ def receive_pull_request_coverage
...
+  end
+
+  def receive_pull_request_total_coverage
```
giordanoluzardo, line 20: 👍

giordanoluzardo (approved):
LGTM @camillof!

PR #156, opened 2025-07-08
Enable sending additional commit status notifying users of the impending migration when sending a commit status

noahd1:
To ensure all customers using pull request integration are aware of the impending migration to qlty.sh, this PR sends an additional failed pull request commit status anytime the pull request service attempts to send a commit status.

The config defaults to false, so in theory, just using updating to the latest version of this gem should be behavior neutral.


noahd1:
FYI, CI is broken, but the tests are passing locally.

